PR #5, opened 2013-07-17
Lost error message and exception

raphael-wb:
When Acquire fails, it may return meaningful error message and exception.

Within ExecuteOperation, "readResult.Combine(result);" overwrites this result with the values of "readResult", which in this block is always blank.

Dropping it just gave me back the error I was looking for.

I've reported the bug in the tracking tool: http://www.couchbase.com/issues/browse/NCBC-281


jeffrymorris:
@raphael-wb 

Can you verify if this has already been fixed? If so can you close this PR?

Thanks!

-Jeff

PR #9, opened 2013-10-13
* Fixes memcached client expiration date issue, adds Enum values to memcached status codes

rrmayer:
* Fixes memcached client expiration date issue
  Ref. Issue #96 (https://github.com/enyim/EnyimMemcached/issues/96)
* Updated StatusCodeEnum consistent with the Memcached protocol (and SASL auth extensions).
* Added code throughout to convert int status codes to the enum.  Added NodeNotFound and Unknown status codes, and return when a node cannot be located.
* Removed unused references.
* Note- Use of the methods that provide memcached feedback, their int results are now enums. This will require very minor code changes, but these changes will improve existing code readability and should be considered beneficial.


jeffrymorris:
@rrmayer -

I believe this fixes the date issue: https://github.com/couchbase/couchbase-net-client/commit/11dd55bfd9c23db18fce4c236ee9f896d8d05faf

You should be able to pull and verify fairly easily. Since there are several changes in a single commit, it's difficult for me to tell what is what TBH. If you could close this PR and then break commit into several smaller ones (based off the release13 branch) I could reassess. 

Thanks!

-Jeff

PR #26, opened 2014-10-31
Remove log events from object finalizers

Svizel:
Order of finalized objects is not guaranteed during application domain
shutdown. Trying to access another object in finalizer will often result
in an exception, when the accessed object was already finallized.

http://msdn.microsoft.com/en-us/library/system.object.finalize%28v=vs.110%29.aspx

### DIFF
--- a/Src/Couchbase/Cluster.cs
+++ b/Src/Couchbase/Cluster.cs
@@ -175,6 +175,7 @@ namespace Couchbase
                 {
                     _clusterController.Dispose();
                 }
+                _disposed = true;
             }
         }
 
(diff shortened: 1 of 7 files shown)


jeffrymorris:
@Svizel -

Per our discussion, I ended up making the finalizers DEBUG only, thus this PR can be closed. Here is the commit that was merged: https://github.com/couchbase/couchbase-net-client/commit/607c2ef4bd73bccb0b8bf57cab2c925fc2b00b94

Thanks a lot for the feedback and PR!

-Jeff

PR #28, opened 2014-11-24
Update IViewClient.cs

thrane:
It would be nice to be able to take the client as a parameter so a an object that needs to use it, so it can be initialized by an IOC container.


jeffrymorris:
@thrane 

Can you please close this PR?

Thanks!

-Jeff

PR #29, opened 2014-12-12
Update ViewRow.cs

Lennartos:
NCBC-722 changed the result of a view so it is returned as ViewRow<T>, which is fine - BUT 
the Key here is returned as object instead of dynamic - making it a hassle to first unbox to dynamic/Jarray to be able to deserialize it / cast it to JValue and then convert. ( which make this unnecessary verbose and cluttered solution ) 
We need to use Key due to mapreduce - so a better solution is to bring it back as dynamic, like it was before.

### DIFF
--- a/Src/Couchbase/Views/ViewRow.cs
+++ b/Src/Couchbase/Views/ViewRow.cs
@@ -18,7 +18,7 @@ namespace Couchbase.Views
         /// The key emitted by the View Map function
         /// </summary>
         [JsonProperty("key")]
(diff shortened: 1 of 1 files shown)


jeffrymorris:
@Lennartos 

This is merged and will be in 2.0.2: https://github.com/couchbase/couchbase-net-client/commit/a7abebd68f9db022abb14731814dea13e81d3a8f

Thanks for the contribution, can you please close this PR?

-Jeff

PR #30, opened 2014-12-17
Both properties should be defined as string 

jelical:
Both properties should be defined as string due to possible verbose 'undefined'  value from server


jelical:
will do it on the fresh fork

PR #33, opened 2015-01-28
update cluster manager interface for Orleans

weitaolee:
https://github.com/dotnet/orleans

when use couchbase cluser manager in Orleans's Grain
for reason-->https://github.com/dotnet/orleans/wiki/External-Tasks-and-Grains

when use HttpClient.PostAsync ,HttpClient.GetAsync ..etc. 
Task.Wait will never schedule by Orleans, so the task  always WaitForActive


weitaolee:
bucket manager has the same problem

weitaolee:
I clone http://review.couchbase.org/ the couchbase-net-client project, but I can not push my code.

please give github developer a easy way to contribute code

PR #34, opened 2015-01-30
update for compatible with Orleans

weitaolee:
https://github.com/dotnet/orleans

when use couchbase cluser manager in Orleans's Grain
for reason-->https://github.com/dotnet/orleans/wiki/External-Tasks-and-Grains

when use HttpClient.PostAsync ,HttpClient.GetAsync ..etc. 
Task.Wait will never schedule by Orleans, so the task always WaitForActive

### DIFF
--- a/Src/Couchbase.Tests.Management/App.config
+++ b/Src/Couchbase.Tests.Management/App.config
@@ -15,7 +15,7 @@
 
   <appSettings>
     <add key="OperationTestAddress" value="127.0.0.1:11210" />
-    <add key="bootstrapUrl" value="http://192.168.62.101:8091"></add>
+    <add key="bootstrapUrl" value="http://192.168.0.100:8091"></add>
     <add key="serverIp" value="192.168.62.101" />
   </appSettings>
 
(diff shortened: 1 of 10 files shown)


jeffrymorris:
@weitaolee -

Thanks for cleaning up the submission! I was going to ask you to do so, but you beat me to it :)

Because this is a change to a previously published interface, this will be scheduled for 2.1.0, which should be around March or April. 

Also, we would prefer that the the synch methods still existed; the async methods that return Tasks should be additional methods added (nothing removed). This would keep any consumers of the API from breaking when these are released. Thoughts?

-Jeff

weitaolee:
Yes,I think you can change the interface by this way.

leave the synch methods, this methods maybe use webrequest to replace httpclient.
and add XXXAsync methods, use the  current implemention.

 I would have done, as you said.Since my time is limited, so I was lazy.
 I will finished this recently.

weitaolee:
@jeffrymorris 
hi, guy
I finished this now. please merge it to master branch.
ClusterManger And BucketManager have both sync and async method now, so couchbase-net-client's customer can update safely.

so this can be scheduled in next release.

jeffrymorris:
@weitaolee -

Once again, thanks for the submission! This is merged and released in 2.0.3, can you close this PR?

Thanks!

-Jeff

PR #36, opened 2015-05-08
Fix: switched values for TcpKeepAlive in ClientConf

dionoid:
Fix: switched around values for TcpKeepAliveTime and
TcpKeepAliveInterval in ClientConfiguration.cs


jeffrymorris:
@dionoid -

This is merged, can you close it?

Thanks for the submission!

-Jeff

PR #37, opened 2015-05-28
Fix connection leak on Memcached bucket dispose

Svizel:
Memcached buckets are using HttpStreamingProvider for Couchbase server
node configuration. This implementation did not properly Dispose
underlying ConnectionPools instances.


jeffrymorris:
@Svizel -

This is now merged, can you close this PR?

Thanks!

-Jeff

PR #38, opened 2015-05-28
Global connection pool setting for app.config

Svizel:
Programmatic configuration allowed setting up connection pools on global level. Add same functionality to configuration through app.config. Also removing DefaultValue=null for optional configuration elements, because null doesn't work anyway.


jeffrymorris:
@Svizel -

Sorry for the delay; this is now in Gerrit awaiting code review. Here is a ticket for tracking: http://issues.couchbase.com/browse/NCBC-919

Thanks for the submission!

-Jeff

PR #39, opened 2015-07-11
Corrected errors when running N1QL queries as bare strings

brantburnett:
Motivation
----------
When running a N1QL query using CouchbaseBucket.Query<T>(string),
QueryClient.Query<T>(Uri, string), or QueryClient.QueryAsync<T>(Uri,
string) the request is not formatted correctly.  N1QL will always return
an error that the statement is blank.

Modifications
-------------
Change QueryClient.Query<T>(Uri, string) and its Async counterpart to
simply format all queries as a QueryRequest and then process through their
counterparts that use a QueryRequest parameter.  For
CouchbaseBucket.Query<T>(string), changed it to use the _requestExecutor
and provide a QueryRequest parameter, just like its Async counterpart that
was already correct.  Updated unit tests to test for these bugs.

### DIFF
--- a/Src/Couchbase.Tests/N1QL/CouchbaseBucket_N1QL_Tests.cs
+++ b/Src/Couchbase.Tests/N1QL/CouchbaseBucket_N1QL_Tests.cs
@@ -51,7 +51,7 @@ namespace Couchbase.Tests.N1QL
         {
             using (var bucket = _cluster.OpenBucket())
             {
-                var query = new QueryRequest("SELECT * FROM `beer-sample` LIMIT 10");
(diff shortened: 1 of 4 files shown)


jeffrymorris:
@brantburnett 

I pushed this to Gerrit our official code review system. It will be reviewed and if it is accepted, will be merged with master and be included in v2.1.4 scheduled for the first week of Aug. 2015.

Thanks for the submission!

-Jeff

jeffrymorris:
@brantburnett 

This is merged; can you please close the ticket?

Thanks,

Jeff

PR #40, opened 2015-07-17
Add option to affect operation flags with value fomatting

Svizel:
The original implementation of ITypeTranscoder does not allow full customization of encoded values. It still binds the encoding by hard wired formatting flags based on the generic type of operation. The new implementation allows for full encoding customization like unwrapping of values passed in as object generic type.


jeffrymorris:
@Svizel - 

I pushed this to gerrit, will review later this week.

Thanks for the submission!

-Jeff

jeffrymorris:
@Svizel -

This is merged; can you please close the PR?

Thanks,

-Jeff

PR #41, opened 2015-07-20
Fix exception when decoding char type

Svizel:
Decode of char type used to lead to invalid cast exception. The exception would flag both save and read operation types as failure. This is now fixed as char is handled separately from string.


jeffrymorris:
@Svizel - 

I pushed this to gerrit, will review later this week.

Thanks for the submission!

-Jeff

jeffrymorris:
@Svizel  -

This is merged; can you please close the PR?

Thanks,

Jeff

PR #47, opened 2015-11-12
Support for more deserialization controls via ITypeSerializers

brantburnett:
Motivation
----------
The current ITypeSerializer implementation works under the assumption that
all serialization/deserialization requests should have the same behavior.
Any client library which implements a custom ITypeSerializer overrides
this behavior for all requests to Couchbase.

However, there are instances where a specific request may require custom
options.  The particular example addressed here is change tracking in the
Linq2Couchbase library.  It needs to control the object creation process
for some deserialization requests in order to create change tracking
proxies.

Additionally, we need a method for custom ITypeSerializer implementations
to provide member name resolution information to consumers.  This will
allow Linq2Couchbase to determine the correct attribute names to use when
building N1QL queries.  Currently, it is forced to assume that the
Newtonsoft.Json behavior is in use.

Finally, there is currently no method to override the deserialization
process for N1QL queries on a per-request basis.

Modifications
-------------
Created a new interface which extends ITypeSerializer named
IExtendedTypeSerializer.

Added GetMemberName method to IExtendedTypeSerializer, which provides
member name resolution information to consumers.

Added DeserializationOptions to IExtendedTypeSerializer, which allows
consumers to set the options they'd like.  Currently, this object supports
only one option, CustomTypeCreator, which allows the consumer to override
the type creation process on a type-by-type basis.

Also provided a SupportedDeserializationOptions object.  This allows the
IExtendedTypeSerializer to define which options it does or does not
support.

Updated the DefaultSerializer to support all of the new interfaces,
methods, and options provided.

Additionally, created a new interface IQueryRequestWithMapper, inherited
form IQueryRequest, which adds a DataMapper to IQueryRequest.  Added this
interface to the default QueryRequest implentation.  This allows the data
mapper used for N1QL queries can be customized on a per-request basis.

Results
-------
For users using the DefaultSerializer based on Newtonsoft.Json, they will
immediately have access to the new features on IExtendedTypeSerializer.
This includes a method to resolve member names, and the ability to
override the type creation process.  This will allow Linq2Couchbase to
transparently implement change tracking proxies.

For projects using a custom ITypeSerializer implementation, consumers such
as Linq2Couchbase can detect support for advanced features by testing for
the IExtendedTypeSerializer interface.  If present, they can then test for
specific features via the SupportedDeserializationOptions property.

Backwards compatibility is fully maintained by these changes.
Additionally, the use of SupportedDeserializationOptions will allow the
addition of more deserialization options in the future without creating
backwards compatibility issues.

### DIFF
--- a/Src/Couchbase.UnitTests/Core/Serialization/DefaultSerializerTests.cs
+++ b/Src/Couchbase.UnitTests/Core/Serialization/DefaultSerializerTests.cs
@@ -1,0 +1,237 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Couchbase.Core.Serialization;
+using Moq;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using NUnit.Framework;
+
+namespace Couchbase.UnitTests.Core.Serialization
+{
+    [TestFixture]
+    public class DefaultSerializerTests
+    {
+        #region DeserializationOptions
+
+        [Test]
+        public void DeserializationOptions_Modification_UpdatesEffectiveSettings()
+        {
+            // Arrange
+
+            var options = new DeserializationOptions();
+            var effectiveSettings = new JsonSerializerSettings();
+
+            var serializer = new Mock<DefaultSerializer>
+            {
+                CallBase = true
+            };
+
+            serializer.Setup(p => p.GetDeserializationSettings(It.IsAny<JsonSerializerSettings>(), options))
+                .Returns(effectiveSettings);
+
+            // Act
+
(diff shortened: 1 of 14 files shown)


brett19:
Thanks for the pull request!!  To ensure quality review, Couchbase employs a [code review system](http://review.couchbase.org/) based on [Gerrit](https://www.gerritcodereview.com/) to manage the workflow of changes in addition to tracking our contributor agreements.

Transferred commit (6189851f03f9865dc091a76312bcb574a1f097b3) to Couchbase Review site:
http://review.couchbase.org/#/c/56960

::SDKBOT/PR:created

jeffrymorris:
@brantburnett -

Thanks Brant, this has been merged! Can you close the PR?

-Jeff

PR #48, opened 2015-11-26
Provide access to cluster from IBucket

brantburnett:
Motivation
----------
Some activites on a bucket might need to access information about the
cluster, such as the number of nodes, cluster status, etc.  Currently,
this information is unavailable from IBucket.

This is based on my own experience creating a set of reactive extensions
for Couchbase:
https://github.com/brantburnett/Rx-Couchbase/commit/99a1480e01f02897f2b404a62dbef11f9e26e31c#diff-d29a87c81f7ee323271660849b15521bR32
In this case it would be helpful to be able to adjust the degrees of
parallelism based on the total number of nodes in the cluster.

This is also based on a conversaion on the forums:
https://forums.couchbase.com/t/cluster-info-from-bucket/5476
In this case, the user was forced to use server info from configuration
only.  They didn't have access to all the servers in the cluster, only the
ones that were configured in advance.

Modifications
-------------
Added a new Cluster property to IBucket to reference the cluster the
bucket was created from.  To access this information easily, the internal
IClusterController interface was modified to include an ICluster property.

Modified the constructors for Cluster to automatically link the
ClusterController to the Cluster when the cluster is created, except in
some unit testing scenarios when the ClusterController is created first.

Also updated ClusterHelper to use the constructor which does this linking
automatically.

Created a test that confirms that the link is in place all the way from
Cluster through to the IBucket implementation.

Results
-------
Both the CouchbaseBucket and MemcachedBucket implementations now provide
easy access to their ICluster.

### DIFF
--- a/Src/Couchbase.Tests/CouchbaseBucketTests.cs
+++ b/Src/Couchbase.Tests/CouchbaseBucketTests.cs
@@ -43,6 +43,15 @@ namespace Couchbase.Tests
         }
 
         [Test]
+        public void Test_GetBucket_HasCluster()
+        {
+            using (var bucket = _cluster.OpenBucket("default"))
+            {
+                Assert.AreEqual(_cluster, bucket.Cluster);
+            }
+        }
+
+        [Test]
         public void When_Key_Does_Not_Exist_Exists_Returns_False()
         {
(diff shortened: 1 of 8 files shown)


brantburnett:
Wasn't recognized by Gerrit, closing to try again

PR #52, opened 2016-01-02
Use default server configuration when there is no bucket config

markvincze:
If in the web.config `<couchbase>` section we only have a `<servers>` section, but no `<buckets>` section (because we don't want to statically configure any buckets), then when we try to open a bucket using its name, the client tries to use localhost:8091 instead of the server configured `<servers>`.  
The scenario is when our config looks like the following

    <couchbaseClients>
      <couchbase useSsl="false" operationLifespan="1000">
        <servers>
          <add uri="http://1.2.3.4:8091/pools"></add>
        </servers>
      </couchbase>
    </couchbaseClients>

This causes two problems.

- If we have no Couchbase server running at localhost:8091, opening the bucket can take seconds, because the client is trying to connect to localhost at first, and only after that fails does it connect to the configured server. This problem and a workaround is described in detail here: http://stackoverflow.com/questions/34183342/why-can-opening-a-couchbase-bucket-be-very-slow.
- If we actually do have a server at localhost:8091, then that is getting used instead of the one configured in `<servers>`. This is clearly a bug.

With this PR, when we get the configuration for a bucket, the server configuration is getting used even if there is no `<buckets>` section.

### DIFF
--- a/Src/Couchbase/Configuration/Server/Providers/ConfigProviderBase.cs
+++ b/Src/Couchbase/Configuration/Server/Providers/ConfigProviderBase.cs
@@ -106,7 +106,10 @@ namespace Couchbase.Configuration.Server.Providers
                 {
                     bucketConfiguration = new BucketConfiguration
                     {
-                        BucketName = bucketName
+                        BucketName = bucketName,
+                        PoolConfiguration = ClientConfig.PoolConfiguration,
+                        Servers = ClientConfig.Servers,
+                        UseSsl = ClientConfig.UseSsl
                     };
                 }
                 else


jeffrymorris:
@markvincze -

Thanks for the pull request!! To ensure quality review, Couchbase employs a [code review system](http://review.couchbase.org) based on Gerrit to manage the workflow of changes in addition to tracking our contributor agreements.

To get this change in and collaborate in code review, please register on Gerrit and accept our CLA. More detailed instructions are available here: http://developer.couchbase.com/open-source-projects#how-to-contribute-code.

Thanks,

Jeff

markvincze:
Hi @jeffrymorris,

I've signed the agreement (I created a new Yahoo account for OpenId, mrk.vincze@yahoo.com).
Is there anything else I have to do?

Cheers,
Mark

jeffrymorris:
@markvincze -

I'll push the PR to Gerrit and review it sometime in the near future. Depending upon how the code review goes, I may ask you to do another patchset, otherwise I'll merge it to master.

-Jeff

jeffrymorris:
@markvincze -

It looks like your email is missing from your gerrit profile can you check this and if it is missing, please add it? I can't push the PR without it (gerrit won't allow it).

Thanks!

-Jeff

markvincze:
Hey @jeffrymorris,

I see email address in two places in the Settings: in Profile, but I can't find where I can edit it.  
The other one is in Contact Information, I added my email address there. Is that the one needed?

Cheers,
Mark

jeffrymorris:
@markvincze - I don't think so, the profile email is what we need.

@ingenthr - can you lend a hand here? I still can't push the PR because of the missing email issue. 

Thanks!

Jeff

ingenthr:
This may be related to the OpenID provider you used when you signed up.  The gerrit review side seems to not have the email address.  @markvincze, do you recall if you allowed sharing of email or not?

Sorry for the trouble here.  Thanks for working through it with us.  I'll look a level deeper on this.  I may also force through the PR in the mean time.

markvincze:
Hey @ingenthr,

No problem. I'm having a bit of a hard time navigating in the UI of gerrit, so maybe I'm missing something :)

I created an OpenId with Yahoo

image

I don't remember having the option to share the email-address during the process.

Now if I access the Settings page in gerrit, I actually see my email address there:

image

Did you manage to set it somehow in the meantime? Yesterday that field was empty.

Cheers,
Mark

jeffrymorris:
@markvincze -

This PR has been pushed to Gerrit successfully; i'll be reviewing it in the next couple of days and it should be in v2.3.0 or 2.2.5 whichever comes first.

Thanks!

-Jeff

jeffrymorris:
@markvincze -

This has been merged with master, can you please close this pull request?

Thanks for the submission!

-Jeff

markvincze:
@jeffrymorris,

Cheers! :)

PR #53, opened 2016-01-07
Improve Timeout Handling of Async Operations

iwillspeak:
If value eviction is turned on in the server some operations don't seem to ever get a response. This is fine when using the client library synchronously, but if you use the async operations they will never time out. This results in a deadlocked task being leaked which takes up one of the connections in the connection pool. After a while no connections are left in the connection pool and operations will fail.

These commits attempt to address that by forcing a the request to time out if no response is received. This is done by starting a timer when each request is sent off for the same length of time as the synchronous timeout, if the timer fires before we get a response the task is cancelled.

This is still a bit rough around the edges, but it seems to solve the issue that I was seeing. Any comments are welcome.

### DIFF
--- a/Src/Couchbase/IO/Connection.cs
+++ b/Src/Couchbase/IO/Connection.cs
@@ -1,9 +1,8 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
-using Common.Logging.Factory;
 using Couchbase.IO.Converters;
 using Couchbase.IO.Utils;
 using Couchbase.Utils;
@@ -15,7 +14,7 @@ namespace Couchbase.IO
         private readonly SocketAsyncEventArgs _eventArgs;
         private readonly AutoResetEvent _requestCompleted = new AutoResetEvent(false);
         private readonly BufferAllocator _allocator;
-        private readonly int Offset;
+        private Timer _asyncOperationTimer;
 
         public Connection(IConnectionPool connectionPool, Socket socket, IByteConverter converter, BufferAllocator allocator)
             : base(socket, converter)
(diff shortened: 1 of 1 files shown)


jeffrymorris:
@iwillspeak -

The email associated with this PR, does not match the email (lith*******e@gmail.com) you have configured in gerrit. I was able to push your older PR because that email (********@crispthinking.com) matches your gerrit account, but this one was rejected. Can you resubmit with your previous email or update you profile in gerrit?

Sorry for the trouble, gerrit and github don't always work friendly together ;)

Thanks!

-Jeff

iwillspeak:
I've added my gmail email to Gerrit now.

Jeffry Morris wrote:

jeffrymorris:
@iwillspeak -

I ended up rebase/squashing your commits into a single change-set (http://review.couchbase.org/#/c/58415/). I'll be reviewing for possible merge for 2.3.0 or 2.2.4 (2.2.3 is locked atm), whichever release comes first, sometime in the next couple weeks.

As for the "rough around the edges", in Gerrit you can easily amend patch-sets to change-set, so as we go through the CR process we can smooth them out ;). LMK if you need any help, but it goes something like this:

git pull gerrit changeset
//change things
git add . //add modified files

git commit --amend
git push gerrit HEAD:refs/for/master 

As a side note, check out the MultiplexedConnection class [here](https://github.com/couchbase/couchbase-net-client/commit/26d142e4cb74ef6d75626c31e37c95b8f9f1522e). It solves the problem you fixed in Connection. 

Thanks!

-Jeff

iwillspeak:
Thanks. The multiplexed connection stuff looks good. Looks like it should speed up heavy write loads a bit by reducing contention on the connection pool quite nicely.

jeffrymorris:
@iwillspeak -

Were still testing this for possible release in 2.2.6 or 2.30 whichever comes first.

Thanks,

Jeff

iwillspeak:
I have already signed the CLA

jeffrymorris:
@iwillspeak -

Sorry, we re-enabled the couchbase bot and it auto-sent that notification. 

Thanks,

Jeff

PR #55, opened 2016-01-31
NCBC-1074: Prevent MVC deadlock on async view queries

brantburnett:
Motivation
----------
When working within MVC actions, executing an asynchronous view query and
then waiting for it to complete synchronously causes a deadlock.

Modifications
-------------
CouchbaseRequestExecutor for aysnchronous view queries needed an
additional call to ContinueOnAnyContext before an await.

Also created an integration test which successfully detects this problem
if it occurs in the future.

Results
-------
MVC actions no longer deadlock in this scenario.


jeffrymorris:
Brant -

This is merged, can you please close this PR?

Thanks,

Jeff

PR #56, opened 2016-03-01
Correct issue in bucket reference counting

eudyj:
When giving out a reference to a cached bucket, the ClusterController was not adding a new reference to the bucket.

However, simply adding a new reference would not be sufficient. Disposing of the bucket is handled by the bucket itself, and while the bucket does call DestroyBucket on its associated ClusterController, thus removing the bucket from the cache, there is no shared locking mechanism. So it is entirely possible that in the ClusterController, between the ContainsKey call and the _buckets[bucketName] call, the bucket could become disposed.

Therefore, AddRef was changed to check the _disposed value and return -1 if the bucket is disposed. It can be used as an atomic add-or-fail operation by the ClusterController. Once the ClusterController finds a bucket in its cache, it will attempt to add the reference before returning the cached bucket. If the add reference fails, it will bootstrap a new bucket.


Src/Couchbase.Tests/Core/ClusterControllerTests.cs
```diff
@@ -178,6 +178,35 @@ public void When_Same_Bucket_Requested_Twice_In_Parallel_Only_One_Bootstrap_Is_D
...
+            using (bucket2 = cluster2.OpenBucket("default", ""))
+            {
+                Thread.Sleep(100); // Sleep while thread1 disposes the bucket
+                await bucket2.ExistsAsync("Key"); // Used to throw ObjectDisposedException
+            }
+
+            t1.Join();
+
+            Assert.AreSame(bucket1, bucket2);
+        }
```
eudyj: In the previous version of the code, this unit test will throw an ObjectDisposedException when one thread disposes the bucket another thread has been given.

jeffrymorris:
@eudyj -

Thanks for the submission! To ensure quality review, Couchbase employs a code review system based on Gerrit to manage the workflow of changes in addition to tracking our contributor agreements.

To get this change in and collaborate in code review, please register on Gerrit and accept our CLA. More detailed instructions are available here: [](http://developer.couchbase.com/open-source-projects#how-to-contribute-code.)

Once you have done this I will push the PR to Gerrit where it will be reviewed.

Thanks,

Jeff

jeffrymorris:
Hi @eudyj -

Have you signed the CLA yet? Let me know if you need any help.

Thanks,

Jeff

jeffrymorris:
@eudyj -

Just checking in with you on signing the CLA, can you do that so we can merge this?

Thanks,

Jeff

PR #58, opened 2016-05-20
Correct issue in bucket reference counting

eudyj:
When giving out a reference to a cached bucket, the ClusterController was not adding a new reference to the bucket.

However, simply adding a new reference would not be sufficient. Disposing of the bucket is handled by the bucket itself, and while the bucket does call DestroyBucket on its associated ClusterController, thus removing the bucket from the cache, there is no shared locking mechanism. So it is entirely possible that in the ClusterController, between the ContainsKey call and the _buckets[bucketName] call, the bucket could become disposed.

Therefore, AddRef was changed to check the _disposed value and return -1 if the bucket is disposed. It can be used as an atomic add-or-fail operation by the ClusterController. Once the ClusterController finds a bucket in its cache, it will attempt to add the reference before returning the cached bucket. If the add reference fails, it will bootstrap a new bucket.


Src/Couchbase.Tests/Core/ClusterControllerTests.cs
```diff
@@ -178,6 +178,35 @@ public void When_Same_Bucket_Requested_Twice_In_Parallel_Only_One_Bootstrap_Is_D
             Assert.IsNotNull(bucket1);
             Assert.AreSame(bucket1, bucket2);
         }
+
+        [Test]
+        public async Task When_Disposing_Bucket_In_Parallel_Does_Not_Dispose_Referenced_Bucket()
```
eudyj: In the previous version of the code, this unit test will throw an ObjectDisposedException when one thread disposes the bucket another thread has been given.

ingenthr:
Hey @eudyj, I see you did the CLA over on review.couchbase.org, but it looks like your email address wasn't associated.  Can you head to http://review.couchbase.org/#/settings/contact and add your email address?  Apologies in advance for any rough edges-- we're working to simplify the CLA verification and code review.  After that we should be able to review/merge the change.

eudyj:
@ingenthr I'm attempting to get my email address into Gerrit but currently when I try to follow the confirmation link, I'm receiving the error "Server error - invalid token". I'll continue to try.

ingenthr:
Interesting.  What authentication provider are you using?  Github OAuth, or one of the OpenID providers?    @brett19 might be able to help debug here as well.

eudyj:
I'm using Github OAuth.

eudyj:
I've tried a couple of different email addresses and browsers, and even signed up using Yahoo instead of Github OAuth, and I keep getting the same message. "Server error - invalid token".

eudyj:
@ingenthr I was finally able to correct this by using a different computer on a different network. It must be a software configuration or firewall issue that was preventing me from confirming my email.

ingenthr:
Ah, great @eudyj.  Apologies for my slow response there, it was a holiday weekend for me.  Glad we got it worked out!  @jeffrymorris will take it from here.

jeffrymorris:
Jason -

This has been merged, can you please close the PR?

Thanks!

-Jeff

PR #63, opened 2016-09-01
Honor connectionTimeout configuration setting when initializing streaming provider

EricCarrGH:
Motivation
----------
I'm new to couchbase. Our dev server crashed and was having a hard time figuring out why initially opening a bucket took 20+ seconds to time out even though I had set the connectionTimeout to 1 second. Stepped through the code and discovered that one of the providers was not setting a timeout when downloading config files from the server.

When connecting to a non responsive server (e.g. cluster.OpenBucket("my-bucket") ), the Streaming Provider does not honor the connectionTimeout property from the config. The Carrier Publication provider correctly honors the configuration. My thoughts are that all providers should honor the bucket config. This hack worked for me, but I do not know enough about the client to know if there is a better way to go about it.

Modifications
-------------
Override the Streaming Provider's HttpClient's Timeout property from the bucket's pool configuration.

Results
-------
Opening a bucket now honors the connectionTimeout bucket pool configuration property.

### DIFF
--- a/Src/Couchbase/Configuration/Server/Providers/Streaming/HttpServerConfig.cs
+++ b/Src/Couchbase/Configuration/Server/Providers/Streaming/HttpServerConfig.cs
@@ -37,6 +37,7 @@ namespace Couchbase.Configuration.Server.Providers.Streaming
             _bucketName = bucketName;
 
             _httpClient = new CouchbaseHttpClient(bucketName, password);
+            _httpClient.Timeout = TimeSpan.FromMilliseconds(clientConfig.BucketConfigs[bucketName].PoolConfiguration.ConnectTimeout);
         }
 
         public string BucketName


jeffrymorris:
@EricCarrGH -

Not sure what's going on. Can you log into Gerrit and look at My/Changes and see if there? If it is, add Jeffry Morris as a reviewer.

-Jeff

jeffrymorris:
@EricCarrGH -

Can you verify that your github email and your gerrit email match? That might be the issue.

-Jeff

jeffrymorris:
@EricCarrGH -

Your commit was done under your other email ***.***@q***l.com...you need to add that as a linked identity in Gerrit. After you do that we should be fine.

Thanks,

Jeff

EricCarrGH:
Ahhh, makes sense. Thank you. I committed using my corp id by mistake. I'll create another commit/PR with my github email so everything matches up.

jeffrymorris:
Got it, thanks!

PR #64, opened 2016-09-01
Honor connectionTimeout configuration setting when initializing streaming provider

EricCarrGH:
Motivation
----------
Our dev server crashed and was having a hard time figuring out why initially opening a bucket took 20+ seconds to time out even though I had set the connectionTimeout to 1 second. Stepped through the code and discovered that one of the providers was not setting a timeout when downloading config files from the server.

When connecting to a non responsive server (e.g. cluster.OpenBucket("my-bucket") ), the Streaming Provider does not honor the connectionTimeout property from the config. The carrierPublication provider correctly honors the configuration. My thoughts are that all providers should honor the config. This hack worked for me, but I do not know enough about the client to know if there is a better way to go about it.

Modifications
-------------
Override the HttpClient's Timeout property using the bucket pool configuration.

Results
-------
Opening a bucket now honors the connectionTimeout bucket pool configuration property.

### DIFF
--- a/Src/Couchbase/Configuration/Server/Providers/Streaming/HttpServerConfig.cs
+++ b/Src/Couchbase/Configuration/Server/Providers/Streaming/HttpServerConfig.cs
@@ -37,6 +37,7 @@ namespace Couchbase.Configuration.Server.Providers.Streaming
             _bucketName = bucketName;
 
             _httpClient = new CouchbaseHttpClient(bucketName, password);
+            _httpClient.Timeout = TimeSpan.FromMilliseconds(clientConfig.BucketConfigs[bucketName].PoolConfiguration.ConnectTimeout);
         }
 
         public string BucketName


jeffrymorris:
@EricCarrGH -

After merging, we found a few issues with your PR. The ticket that was raised is here: https://issues.couchbase.com/browse/NCBC-1193

At this point you can resubmit a new PR with suggestions in the ticket or the ticket will go into the queue and picked up by another developer internally for 2.3.7. 

Thanks,

Jeff

EricCarrGH:
Jeff, after reviewing the notes in NCBC-1193, I can proceed with creating a new top level ClientConfiguration.BootStrapTimeout for this scenario, with tests, if that is your preference.

jeffrymorris:
@EricCarrGH -

Great, sounds good! With Gerrit you can do `git commit --amend` and push each patchset...this makes it easy to iterate through a series of code reviews. 

-Jeff

jeffrymorris:
@EricCarrGH - 

The default timeout for N1QL, Views and FTS is 75000ms...we can start with that. I'll check with the other SDK's to see what they are doing, in general we want parity.

If you look at the QueryRequestTimeout and ViewRequestTimeout properties, you'll see what needs to be done to ensure that the BootstrapTimeout's values are propagated correctly for JSON config, standard App.Config and for programmatic configuration.

-Jeff

EricCarrGH:
@jeffrymorris - Sounds good. Thanks for the info. I will proceed and follow up if I have any more Qs.

-Eric

EricCarrGH:
FYI - I've had a busy week and now will be traveling on vacation through Oct 1 - if nobody picks this up while I'm away I'll work on it after I return.

-Eric

jeffrymorris:
@EricCarrGH -

Sounds good. Enjoy your vacation!

-Jeff

PR #71, opened 2017-04-20
Avoid blocking in async code in request executers

pkese:
I've noticed some blocking Tread.Sleep calls being made inside async code.  

In addition `async Task SendAsync(IOperation operation)` was being called without awaiting on results.

That may as well be intentional, however it seems to be a good practice to await on the result in order for exceptions to be handled normally (e.g. unwrapped from `AggregateException`) by an `await` after calling async functions.

### DIFF
--- a/Src/Couchbase/Core/Buckets/CouchbaseRequestExecuter.cs
+++ b/Src/Couchbase/Core/Buckets/CouchbaseRequestExecuter.cs
@@ -713,7 +713,7 @@ namespace Couchbase.Core.Buckets
         /// <returns>
         /// An <see cref="Task{IOperationResult}" /> object representing the asynchronous operation.
         /// </returns>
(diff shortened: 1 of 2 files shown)


jeffrymorris:
@pkese -

Thanks for the submission, @MikeGoldsmith has some comments available for you in the changeset in Gerrit. To create a a new patchset:

- Add Gerrit as remote: $ git remote add gerrit ssh://<your-gerrit-user-name>@review.couchbase.org:29418/couchbase-net-client.git

- Do a remote update: $git remote update

- Make your changes

- Add them to the current changeset: $git add .
- Do a commit amend: $git commit --amend
-Push the changes to gerrit: 
$ git push gerrit HEAD:refs/for/master

If you run into any issues, let Mike or myself know and we can assist. 

Thanks,

Jeff

PR #74, opened 2017-08-16
Deserialize QueryPlan with DefaultSerializer

ross-stovall:
QueryPlan won't deserialize properly if the Newtonsoft serializer is not used. This pull request ensures QueryPlan's are always deserialized using the DefaultSerializer.

### DIFF
--- a/Src/Couchbase/N1QL/QueryClient.cs
+++ b/Src/Couchbase/N1QL/QueryClient.cs
@@ -84,6 +84,7 @@ namespace Couchbase.N1QL
             }
             var query = new QueryRequest(statement);
             query.BaseUri(toPrepare.GetBaseUri());
+            query.DataMapper = new QueryDataMapper();
             return ExecuteQuery<QueryPlan>(query);
         }
 
@@ -108,6 +109,7 @@ namespace Couchbase.N1QL
             }
             var query = new QueryRequest(statement);
             query.BaseUri(toPrepare.GetBaseUri());
+            query.DataMapper = new QueryDataMapper();
             return await ExecuteQueryAsync<QueryPlan>(query, cancellationToken).ContinueOnAnyContext();
         }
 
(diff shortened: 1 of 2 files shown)


jeffrymorris:
Thanks @ross-stovall!

MikeGoldsmith:
Hi @ross-stovall 

Thanks for the contribution.

I'm not sure we need a custom QueryPlanSerializer type, instead we could assign a DefaultSerialzier instance specifically for use with Query Plans. This could also be instantiated during the QueryClient constructor to be reused as needed.

What do you think?

ross-stovall:
Hey there @MikeGoldsmith 

Just invoking the DefaultSerializer was my initial idea but I couldn't find where to make that happen as the Query Plan query is executed as a generic QueryRequest. Is there a way to define the serializer using the ExecuteQuery method without defining the DataMapper? Or is there already an existing default DataMapper we should be using?

MikeGoldsmith:
Hey @ross-stovall 

Ah, I see the problem. JsonDataMapper requires a ClientConfiguration parameter, which is not necessary. If it's constructor was changed to take the ITypeSerializer directly it would be easier to instantiate.

Are you ok with me making an amend I'll show you want I mean?

ross-stovall:
Hey @MikeGoldsmith 

Yeah when I looked at JsonDataMapper it looked like the intention was to couple it with the passed in ClientConfiguration so I intentionally didn't want to mess around in there. But adding a constructor to directly set the ITypeSerializer would definitely simplify things. I'm good with you making the change.

Thanks much.

Ross

MikeGoldsmith:
Hi @ross-stovall 

I've pushed an update to our Gerrit system, you can see it here:
http://review.couchbase.org/#/c/82340/

I also updated the commit message to follow our preferred format, hope that's ok. Let me know wha t you think.

Thanks again.

ross-stovall:
Hey @MikeGoldsmith 

Sorry for the little bit of a late reply. We just had a huge typhoon hit us the past couple days. Your change looks good to me. Reusing the same instance is much better. Let's ship it ;).

Best.

Ross

PR #79, opened 2017-10-05
NCBC-1549: Add async creation of buckets.

dlemstra:
Motivation
----------
Buckets cannot be created asynchronously at the moment and being able
to do this could be a big win under a high load.

Modifications
-------------
 - Added OpenBucketAsync to ICluster
 - Added CreateBucketAsync to IClusterController
 - Added unit tests for the OpenBucket methods of Cluster.
 - Added GetBucketAsync to ClusterHelper and changed the locking logic in that class.

Result
------
Buckets can now be created asynchronously :)

### DIFF
--- a/Src/Couchbase.UnitTests/ClusterTests.cs
+++ b/Src/Couchbase.UnitTests/ClusterTests.cs
@@ -1,0 +1,159 @@
+﻿using Couchbase.Authentication;
+using Couchbase.Configuration.Client;
+using Couchbase.Core;
+using Moq;
+using NUnit.Framework;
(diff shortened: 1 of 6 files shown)


MikeGoldsmith:
Hi @dlemstra - sorry for the slow reply, we've been caught up with conferences the last week or so. We'll take a look at your update in the next day or so.

Thanks

dlemstra:
@MikeGoldsmith No problem 👍 Would love to start with the rest of the async improvments when this PR is merged. This is only the first step.

PR #81, opened 2017-11-21
Support for controlling how MemoryStreams are created

djluck:
# Summary
This PR adds the ability to configure how `MemoryStream` instances are created within the Couchbase SDK, while maintaining compatibility with how `MemoryStreams` instances are created.

# Motivation
When profiling our production web servers, I have observed that a large amount of allocated memory can be traced back to the Couchbase SDK and it's use of `MemoryStream`:
image

This change would allow users to plug in their own more memory efficient `MemoryStream` implementations (see [Microsoft.IO.RecyclableMemoryStream](https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream)).

# Changes
- Added the `MemoryStreamFactory` that can use a configured function to create MemoryStream instances
- Default implementation does not differ from current, `new MemoryStream()` is called
- In instances where `new MemoryStream()` was being called, replaced with `MemoryStreamFactory.GetMemoryStream()`
- In instances where `new MemoryStream(byte[])` was being called, no changes were made: MemoryStreams are being used in a read-only capacity.

# Notes
This is my first PR against this project- if my code does not meet your standards or the static factory is not a pattern you wish to encourage, please let me know and I'll do my best to re-work this PR.

### DIFF
--- a/Src/Couchbase/IO/MemoryStreamFactory.cs
+++ b/Src/Couchbase/IO/MemoryStreamFactory.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using System.IO;
+
+namespace Couchbase.IO
+{
+    /// <summary>
+    /// Creates instances of MemoryStreams for use in writing/ reading bytes to/ from the network.  
+    /// </summary>
+    public static class MemoryStreamFactory
+    {
+        private static Func<MemoryStream> FactoryFunc = () => new MemoryStream();
+        
+        /// <summary>
+        /// Provides a custom MemoryStream creation function that will override the default implementation.
+        /// </summary>
+        /// <param name="factoryFunc"></param>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static void SetFactoryFunc(Func<MemoryStream> factoryFunc)
+        {
+            if (factoryFunc == null)
+            {
+                throw new ArgumentNullException(nameof(factoryFunc), "You must provide a non-null factory function");
+            }
+
+            FactoryFunc = factoryFunc;
+        }
+        
+        /// <summary>
+        /// Fetches a MemoryStream. The default implementation retuns a new MemoryStream instance.
+        /// </summary>
+        /// <returns></returns>
+        public static MemoryStream GetMemoryStream()
+        {
+            return FactoryFunc();
+        }
+    }
+}
--- a/Src/Couchbase/Core/Serialization/DefaultSerializer.cs
+++ b/Src/Couchbase/Core/Serialization/DefaultSerializer.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using Couchbase.IO;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 
@@ -168,7 +169,7 @@ namespace Couchbase.Core.Serialization
         /// <returns>A <see cref="byte"/> array that is the serialized value of the key.</returns>
         public byte[] Serialize(object obj)
         {
-            using (var ms = new MemoryStream())
+            using (var ms = MemoryStreamFactory.GetMemoryStream())
             {
                 using (var sw = new StreamWriter(ms))
                 {
--- a/Src/Couchbase/Core/Transcoders/BinaryTranscoder.cs
+++ b/Src/Couchbase/Core/Transcoders/BinaryTranscoder.cs
@@ -3,6 +3,7 @@ using System;
 using System.IO;
 using System.Runtime.Serialization.Formatters.Binary;
 using Couchbase.Core.Serialization;
+using Couchbase.IO;
 using Couchbase.IO.Converters;
 
 namespace Couchbase.Core.Transcoders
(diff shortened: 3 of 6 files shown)


Src/Couchbase/Couchbase.csproj
```diff
@@ -1,61 +1,53 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
-
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
MikeGoldsmith: Are there any changes in the csproj other than formatting? If just formatting, can we ignore them for this commit?

Src/Couchbase/IO/MemoryStreamFactory.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.IO;
+
+namespace Couchbase.IO
+{
+    /// <summary>
+    /// Creates instances of MemoryStreams for use in writing/ reading bytes to/ from the network.  
+    /// </summary>
+    public static class MemoryStreamFactory
```
MikeGoldsmith: I think it would be good if we could extend this class to be able to setup a custom MemoryStream using configuration (app/web.config) too.

Src/Couchbase/Couchbase.csproj
djluck: Sure, I'll remove this change

MikeGoldsmith:
I've created issue [NCBC-1587](NCBC-1587) to track the change in JIRA.

Src/Couchbase/IO/MemoryStreamFactory.cs
djluck: Yeah, making this configurable would be a good idea but how would you envision this working? In it's current state, the only way to provide a custom `MemoryStream` is via a delegate. Would the config value accept a static method in the user's code base that was then called via reflection?

An alternative is to expand the scope of this PR and include the Microsoft.IO.RecyclableMemoryStream package and then expose the various configuration knobs to the user (this was the main reason I didn't include the package as part of the PR- finding good default values for the size of the `byte[]` blocks allocated and the maximum size of the pool would be use-case dependent). What do you think?

djluck:
@MikeGoldsmith thanks for the quick feedback. Once you provide the JIRA number, I'll fix up the commit message and the formatting. I think further discussion about the configuration would be required and I'm happy to put the work in.

Src/Couchbase/IO/MemoryStreamFactory.cs
MikeGoldsmith: I'd go for putting the MemoryStream type into config then during init create an instance of that type to use in the factory. It's a little bit of reflection but should only happen on start-up. 

We would only get the default values so if a user wanted to tweak specific implementation settings they would have to do it programatically. This way we support any implementation of MemoryStream without locking ourselves into a particular one

djluck: The issue with providing the name of a type in the config is that object pools like Microsoft.IO.RecyclableMemoryStream require you to create `MemoryStream` instances using a central factory- you can't new up the `MemoryStream` implemented by this package. Or have I misunderstood you, did you mean that the config could accept the name of a method?

MikeGoldsmith: Ah, the Microsoft.IORecyclableMemoryStream doesn't have a parameterless constructor like System.IO.MemoryStream. That would be tougher to manage via configuration.

I think it's best we leave the configuration piece for now and we can come back to it.

djluck: Ok that might be for the best. Thanks for the discussion man, most welcome 👍

djluck:
One more thing @MikeGoldsmith- if I do a force push to fix the commit message and formatting issues, will this mess with Gerrit / your overall process?

MikeGoldsmith:
I think if you force push changes to this PR, it should create a new patch on the same changeset in Gerrit - which is what we want.

jeffrymorris:
@djluck - Very cool, thanks for submitting!

djluck:
@MikeGoldsmith pushed, let me know if you need anything else from me and thanks for the review 👍

MikeGoldsmith:
@djluck Thanks for the update, I also updated your commit message in Gerrit to wrap and 72 chars and the section titles formatting. No big deal, just something we try to do so it's easy to read via `git log`.

I have +2'd your change and would like to get a review from @jeffrymorris too. He's based in the US so may have to wait until the beginning of next week due to Thanks Giving national holiday. I expect it to be in the 2.5.3 release scheduled for 5th Dec 2017.

Thanks again.

djluck:
That's awesome, thanks @MikeGoldsmith

MikeGoldsmith:
Hi @djluck - @jeffrymorris has reviewed and has a suggestion. You can see his review comments on Gerrit [here](http://review.couchbase.org/c/85863).

Let me know if you have any trouble viewing. Thanks.

MikeGoldsmith:
Hi @djluck - I'd really like to get this change into the next release (2.5.3 scheduled for 5th Dec) so am happy to pick up the client configuration work Jeff requested.

I'll make a start on it and let you know how it goes.

MikeGoldsmith:
Ignore the above comment, I see you've left a comment directly on the Gerrit change set.

MikeGoldsmith:
@djluck - 2.5.3 has been released and is available on nuget.org with your above change included

djluck:
Thanks for the update mike, looking forward to including this in our next
release.

PR #84, opened 2018-05-30
Fix documentation of Expiry

runerei:
The value is in seconds, not milliseconds.
https://developer.couchbase.com/documentation/server/4.1/developer-guide/expiry.html

### DIFF
--- a/Src/Couchbase/Document.cs
+++ b/Src/Couchbase/Document.cs
@@ -21,7 +21,7 @@ namespace Couchbase
         public ulong Cas { get; set; }
 
         /// <summary>
-        /// The time-to-live or TTL for the document before it's evicted from disk in milliseconds.
+        /// The time-to-live or TTL for the document before it's evicted from disk in seconds.
         /// </summary>
         /// <remarks>Setting this to zero or less will give the document infinite lifetime</remarks>
         public uint Expiry { get; set; }


MikeGoldsmith:
Hi @runerei 

Thanks for the PR. However, the Document.Expiry property is expressed as milliseconds, not seconds. You can see it in action [here](https://github.com/couchbase/couchbase-net-client/blob/master/Src/Couchbase.IntegrationTests/MemcachedBucketTests.cs#L689) (note: this is a Memcached bucket test, but applies to Couchbase buckets too).

Also, I'm unsure why you think the link you provided indicates timeouts are mean to be in seconds, the linked page does not describe the expiry precision and the .NET [example](https://developer.couchbase.com/documentation/server/4.1/sdks/dotnet-2.2/time-to-live.html#concept_ywd_w1b_4t) shows it be used with milliseconds.

If my assumptions are incorrect, or I've misunderstood, please let me know.

Thanks.

runerei:
Sorry if I'm wrong. From this code it looked like it was using seconds.

https://github.com/couchbase/couchbase-net-client/blob/64fdf7225997b77a59b55f6d687482a355538865/Src/Couchbase/Utils/TimeSpanExtensions.cs#L18

However there are some string parsing in the same class implying support for different time resolutions.

What I found was that if the expiry is in milliseconds it will overflow after 50 days.

Does it use milliseconds when expiry is less than 30 days and seconds when it is more?

MikeGoldsmith:
Hmm, it looks like there is some confusion regarding these values.

I'll look into it further and report back. I've created a JIRA ticket [NCBC-1714](https://issues.couchbase.com/browse/NCBC-1714) to track it. Thanks for the report and sorry for the confusion.

runerei:
As a comment to  [NCBC-1714](https://issues.couchbase.com/browse/NCBC-1714), using ```uint``` and not ```TimeSpan``` how will you express expiry of

* 10 seconds
* 10 months

And what is the expected ```expiration``` value of the raw Couchbase document?

By the way, in [this example ](https://github.com/couchbaselabs/devguide-examples/blob/master/nodejs/expiration.js) node.js is using expiry in seconds.

MikeGoldsmith:
Hi @runerei 

I finally got some time to look into this properly and think I've found the points have have been confusing us both;
- document expiries are _always_ expressed as `seconds` if the value is less 30 days or a `unix timestamps` if greater on the server - [link](https://developer.couchbase.com/documentation/server/4.1/developer-guide/expiry.html)
- the .NET API exposes three ways to control the expiry (not ideal):
  - directly in the API via a TimeSpan or uint parameter (seconds)
  - As a uint on the Document class (expressed as milliseconds)

This is where the confusion comes in; we use an internal function [ToTtl()](https://github.com/couchbase/couchbase-net-client/blob/64fdf7225997b77a59b55f6d687482a355538865/Src/Couchbase/Utils/TimeSpanExtensions.cs#L18) to convert either TimeSpan or uint into the expiry value expressed as seconds.

Likely Document.Expiry should not be a uint, and should be a TimeSpan for consistentcy - same for removing the unit variant on the bucket API. For now, we can't change that as it would be a breaking change, hopefully we can do better in the 3.0 API.

PR #89, opened 2018-10-19
Trapping exception from config serialization

smitcham:
The Serialization of the config can through an exception due to
'Input string is not in the proper format'.  When the exception
is thrown the CT thread stops and the _configQueue grows without
bound.


smitcham:
There were no unit tests for this method, and I wasn't sure how to set them up.  Also, I did the most naive fix by wrapping the serialization, you may want to do something different.

smitcham:
closing in favor of internal patch.

PR #90, opened 2019-01-23
Remove the head of the queue when popping

terkill:
QueryPop removed the tail of the queue instead of the head.


MikeGoldsmith:
Hi @terkill 

Thank you for the PR. The change looks good.

I've updated the commit message to follow our standard structure and accepted the change. It will be part of the 2.7.4 release, which is currently scheduled for 5th Feb 2018.

PR #91, opened 2019-03-31
See OperationBase Timedout method:

SunnySimantov:
public bool TimedOut()
        {
            if (_timedOut) return _timedOut;

            var elasped = DateTime.UtcNow.Subtract(CreationTime).TotalMilliseconds;
            if (elasped >= Timeout || (ErrorCode != null && ErrorCode.HasTimedOut(elasped)))
            {
                _timedOut = true;
            }
            return _timedOut;
        }

It compared the elapsed duration in millseconds with the Timeout property - which is set in seconds in the constructor of Get! This is a mismatch which causes wrong OperationTimedout error messages.

### DIFF
--- a/Src/Couchbase/MemcachedBucket.cs
+++ b/Src/Couchbase/MemcachedBucket.cs
@@ -2685,7 +2685,7 @@ namespace Couchbase
         /// </returns>
         public IOperationResult<T> Get<T>(string key, TimeSpan timeout)
         {
-            var operation = new Get<T>(key, null, _transcoder, timeout.GetSeconds());
+            var operation = new Get<T>(key, null, _transcoder, timeout.GetMilliseconds());
             return _requestExecuter.SendWithRetry(operation);
         }
 
--- a/Src/Couchbase/Utils/TimeSpanExtensions.cs
+++ b/Src/Couchbase/Utils/TimeSpanExtensions.cs
@@ -49,6 +49,16 @@ namespace Couchbase.Utils
             return (uint) timeSpan.TotalSeconds;
         }
 
+        /// <summary>
+        /// Retrieves the number of milliseconds expressed in a <see cref="TimeSpan"/> as an <see cref="uint"/>.
+        /// </summary>
+        /// <param name="timeSpan">The timespan.</param>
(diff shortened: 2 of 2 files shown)


jeffrymorris:
Hi @SunnySimantov -

Can you [register on Gerrit](http://review.couchbase.org/) and [accept the CLA](http://review.couchbase.org/#/settings/new-agreement) so this can go through a code review?

Thanks,

Jeff

SunnySimantov:
Hello @jeffrymorris. If all went right, I registered and accepted the CLA.
Please take notice that my fix is only in one point of the code, but the inconsistency is found in other places as well.
Also, you must probably be aware of this:
https://github.com/couchbase/couchbase-net-client/commit/85a3342ced182ef49d033de9deeaa588da2c84fc

This has fixed the time out error messages, but only in an indirect way, as the main issue from what I see, is the comparison of milliseconds and seconds.

jeffrymorris:
@SunnySimantov -

I don't see the PR in Gerrit, please ensure the following:

- Your gerrit account email and your Github PR email match
- You have accepted the CLA in the Gerrit settings page under "Agreements" (old UI only!)

Once that is done I should see it.

-Jeff

jeffrymorris:
@SunnySimantov -

I found an existing [Jira ticket](https://issues.couchbase.com/browse/NCBC-1776) that covers this issue and it already has a commit that fixes all occurences of the bug. It was merged and reverted at some point because it failed testing; we have reopened it and will use this commit over the one you added because of the completeness. Thank you very much for contributing and helping us get the issue resolved!

-Jeff

SunnySimantov:
Hello Jeffry, 

Thank you for updating me about this issue. I indeed accepted the CLA and registered with the correct mail, but seems that it didn't work. Since I'm anyway very busy with me regular day-to-day work, and since the bug was fixed indirectly between 2.7.4 and 2.7.6 (with the IsTimedout method, if I'm correct), I gave up on continuing with the pull request.

I'm happy to see that you are still taking care of the issue, regardless of who is responsible for the commit. What's important is making the SDK better. :)

PR #92, opened 2019-06-11
Proposal to not making this strange strange to the API

sam-fold-cko:
Removing the async suffix from all the async methods is a bad idea... from a Couchbase enterprise customer


jeffrymorris:
@sam-fold-cko -

Thanks for the feedback! Yes, its an idea we have been throwing out there; largely because there is no synchronous API and no real reason to seperate async from sync calls. There is a ticket available for comment: https://issues.couchbase.com/browse/NCBC-1959

Thanks,
Jeff

jeffrymorris:
http://review.couchbase.org/c/110848/

PR #93, opened 2019-08-07
Fix typo ElaspedTime in Metrics

MirekVales:


jeffrymorris:
@MirekVales -

Thanks for submitting. Since this would be a breaking change to a public class, we would have to release in a major release.

-Jeff

MirekVales:
Guys, the CLA signing does not work correctly. I can see "Individual contributor agreement. Signed already." but the robot has closed it anyway due to timeout.

MikeGoldsmith:
Hi @MirekVales - please can you check the same email you used to sign the commit is the same one you use for github please. That's how it verifies the CLA has been signed.

Thanks

PR #94, opened 2019-09-26
Implemented INode to create default method.

WallaceValadao:


MikeGoldsmith:
Hi @WallaceValadao 

Thank you for the contribution and I like what you've done.

My only suggestion is for the const string for _NostPlaceholder_ to be placed in the new Node class so it's easier to find and keep related items together.

WallaceValadao:
Hi @MikeGoldsmith 

Thank for your return.
I'll make this change. I did it because the const string NostPlaceholder  is using in many places in (Couchbase/Core/Configuration/Server).

MikeGoldsmith:
Hi @WallaceValadao 

Just to let you know, you need to sign the CLA before we can accept any contribution. There is a time limit before a PR is closed automatically.

I'd be happy to accept this change, along with the small change discussed above.

Thanks

PR #96, opened 2020-01-09
NCBC-2235 GetResultWithValue now calls GetMultiValues instead of GetValue

mfeerick:
GetResultWithValue was setting result.Value to the result of calling GetValue which returns typeof T instead of the expected IList<OperationSpec> which was throwing an invalid cast exception.

I could not change the GetValue method to return IList<OperationSpec> as is it used in other areas of the code and is actually a base class method. Instead of this I created a new function called GetMultiValues which does the same procedure as GetValue but instead returns the populated _lookupCommands member variable instead which is an IList<OperationSpec> as required.

The result being GetResultWithValue no longer fails and the individual LookupIn path values get be retrieved.


jeffrymorris:
New commit on gerrit/release27: http://review.couchbase.org/c/120248/

PR #97, opened 2020-02-05
Don't export fuzziness or prefix_length when values less than 1

nekstrom:
Currently Couchbase 6.5 does not accept a fuzziness value of 0 while Couchbase 6.0 did.  As it stands the 2.7 SDK cannot make a search for an exact value in FTS.

This stops the code from outputting "fuzziness":0 or "prefix_length":0 allowing the search to succeed.


jeffrymorris:
@nekstrom -

Thanks for the submission!

It looks like you need to create a Gerrit account so that this commit will be pushed to CR. Can you do that using the same email used when pushing the PR?

I created a ticket for tracking: https://issues.couchbase.com/browse/NCBC-2361

-Jeff

PR #103, opened 2020-06-22

smitcham:


smitcham:
I don't know if this one worked.  From things I see on stack overflow about this, my identity needs to be completely nuked from the review.couchbase.org user database in order for me to fix the bad email.

andy-williams:
> I don't know if this one worked. From things I see on stack overflow about this, my identity needs to be completely nuked from the review.couchbase.org user database in order for me to fix the bad email.

I've had the same issue with my previous PR, if you signed up using your github account, a workaround that worked for me is to change your primary github email and then open another (new) PR.

jeffrymorris:
@brett19 - any thoughts as to why we cannot get this into gerrit?

-Jeff

smitcham:
Maybe someone with admin can delete the account 1004380 so that I can start
over with only my gmail address on my github account?

PR #106, opened 2020-07-05

smitcham:
Another attempt with a new auth token that maybe only has my email.

The stream writer is causing the recyclable memory stream to be disposed prior to the ToArray call being made.

### DIFF
--- a/Src/Couchbase.UnitTests/Couchbase.UnitTests.csproj
+++ b/Src/Couchbase.UnitTests/Couchbase.UnitTests.csproj
@@ -31,6 +31,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="1.3.4" />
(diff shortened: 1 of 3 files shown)


smitcham:
The system still thinks my work email is active, so it must be coming from the gerrit account.

jeffrymorris:
@smitcham - did you try to create a completely new profile using a different email and then push a PR using the new email? Are you sure your not using sm****am@bliz****.com to push the PR?

We'll get this worked out!

-Jeff

brett19:
Hey @smitcham,
Looks like your commit was authored by the `sm****am@bliz****.com` email, this is something configured on your local Git client, I've included some information below on how you can change that.  Once you've updated your email, feel free to amend the commit with your updated email and push it back here where it should be pulled directly into Gerrit.

```
$ git config --global user.name "John Doe"
$ git config --global user.email "john@doe.org"
```

Cheers, Brett

PR #107, opened 2020-07-14

smitcham:
I've checked my local commits to ensure that they have the right email on the commit.

### DIFF
--- a/Src/Couchbase.UnitTests/Couchbase.UnitTests.csproj
+++ b/Src/Couchbase.UnitTests/Couchbase.UnitTests.csproj
@@ -31,6 +31,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="1.3.4" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="Moq" Version="4.7.8" />
     <PackageReference Include="NUnit" Version="3.6.1" />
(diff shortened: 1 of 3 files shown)


smitcham:
looks like it has the right email this time.

jeffrymorris:
@smitcham - yeah, email looks right, but I don't see the commit yet in Gerrit. Try doing this:

1. Login to Gerrit
2. Click your Avatar on the upper right and then "Settings"
3. Select "Email Addresses" on the menu to your left and then add a new email using the "New Email Address" Textbox and click "Save"

Once you do this in Gerrit, it should come through.

-Jeff

smitcham:
updated. commenting to see if that triggers things.

smitcham:
review was created but isn't pointing to release27

brantburnett:
@smitcham Can I recommend one tweak to this? One of the advantages of disposing the StreamWriter before calling ToArray is that ensures that the writer and encoder are fully flushed to the stream: https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/IO/StreamWriter.cs#L190. Instead of calling ToArray() before disposing, I'd recommend using the constructor overload on StreamWriter to pass `leaveOpen` as `true`. This way the dispose won't also dispose the underlying Stream, but the flush still occurs.

jeffrymorris:
@smitcham - SDKBOT pushes to the master branch on Gerrit by default. I pushed to release27 and abandoned the commit on master.  Since your now setup on Gerrit, I suggest you just push to Gerrit directly from now on - just add Gerrit as a remote (`git remote add gerrit ssh://[your username]@review.couchbase.org:29418/couchbase-net-client.git ` . Note on Gerrit, the workflow is:

1. `git add .`
2. `git commit --amend`
3. `git push gerrit HEAD:refs/for/release27`

@brantburnett can you add your comment ^ to the review on gerrit?

PR #109, opened 2020-11-01
register provider as scoped

bobbyangers:
When disposing voluntarily (or not);  all subsequent requests are affected.

Furthermore; if a bucket get's disposed; then the cached reference becomes unsuable. The provider should check if the object in his collection is still valid before returning it (GetOrAdd)

### DIFF
--- a/src/Couchbase.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Couchbase.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -67,7 +67,7 @@ namespace Couchbase.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(bucketName));
             }
 
-            services.AddSingleton(serviceProvider =>
+            services.AddScoped(serviceProvider =>
             {
                 var generator = serviceProvider.GetRequiredService<NamedBucketProxyGenerator>();
 
@@ -93,7 +93,7 @@ namespace Couchbase.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(bucketName));
             }
 
-            services.TryAddSingleton(serviceProvider =>
+            services.TryAddScoped(serviceProvider =>
             {
                 var generator = serviceProvider.GetRequiredService<NamedBucketProxyGenerator>();
 
--- a/src/Couchbase.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Couchbase.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -40,8 +40,8 @@ namespace Couchbase.Extensions.DependencyInjection
 
             services.AddSingleton<NamedBucketProxyGenerator>();
             services.TryAddSingleton<ICouchbaseLifetimeService, CouchbaseLifetimeService>();
-            services.TryAddSingleton<IClusterProvider, ClusterProvider>();
-            services.TryAddSingleton<IBucketProvider, BucketProvider>();
+            services.TryAddScoped<IClusterProvider, ClusterProvider>();
+            services.TryAddScoped<IBucketProvider, BucketProvider>();
 
             if (options != null)
             {
--- a/src/Couchbase.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Couchbase.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -40,6 +40,30 @@ namespace Couchbase.Extensions.DependencyInjection
 
             services.AddSingleton<NamedBucketProxyGenerator>();
             services.TryAddSingleton<ICouchbaseLifetimeService, CouchbaseLifetimeService>();
+            services.TryAddSingleton<IClusterProvider, ClusterProvider>();
+            services.TryAddSingleton<IBucketProvider, BucketProvider>();
+
+            if (options != null)
+            {
+                services.Configure(options);
+            }
+
+            return services;
+        }
+
+        /// <summary>
+        /// Add Couchbase dependencies to the <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="options">Optional action to configure the <see cref="ClusterOptions"/>.</param>
+        /// <returns>The <see cref="IServiceCollection"/>.</returns>
+        public static IServiceCollection AddScopedCouchbase(this IServiceCollection services,
+            Action<ClusterOptions> options)
+        {
+            services.AddOptions();
+
+            services.AddSingleton<NamedBucketProxyGenerator>();
+            services.TryAddSingleton<ICouchbaseLifetimeService, CouchbaseLifetimeService>();
             services.TryAddScoped<IClusterProvider, ClusterProvider>();
             services.TryAddScoped<IBucketProvider, BucketProvider>();
 
@@ -67,7 +91,7 @@ namespace Couchbase.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(bucketName));
             }
 
-            services.AddScoped(serviceProvider =>
+            services.TryAddSingleton(serviceProvider =>
             {
                 var generator = serviceProvider.GetRequiredService<NamedBucketProxyGenerator>();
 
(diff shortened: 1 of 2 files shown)


brantburnett:
@bobbyangers 

I'm afraid I must disagree with the content of this PR. For performance reasons, clusters and buckets should be scoped to the lifetime of your application, not the lifetime of a request. If it disposes and recreates for every request then the performance will seriously suffer. Instead, the cluster and bucket should simply not be disposed once received from the provider.

bobbyangers:
I understand that connection pools and such have to be singleton;  And I understand that there is a small performance hit;  but the issue is that disposable object can be disposed....  simply saying they should not is  not "safe" programming.
I also understand that the programming paradigm would make sense in a persistent;  but in a dotnet/web;  all request should be independent;  the outcome of one request should not affect the subsequent requests.  

The whole request pipeline is designed in such.

Keeping around a disposable object in between requests, is not a good idea.

bobbyangers:
Another suggestion then,  why not simply create another extensions  method;  ie   **AddCouchbaseScoped**

And let the user choose which one he needs for his purpose ?

brantburnett:
It's not the objects that are expensive and problematic, it's the connection pool. When a cluster is bootstrapped in the SDK, there are many network operations involved in bootstrapping multiple connections to multiple Couchbase nodes. This is an expensive operation and should only be performed on application start.

The core SDK is designed with these objects as disposable, which closes the connection pool. This is frankly a bit confusing since it's generally meant to be used as a singleton (or, more precisely, a singleton per cluster of Couchbase Servers). But at this point, we're stuck with it. IDisposable can't be removed without breaking semantics for all existing consumers.

I certainly agree that request should be idempotent, but singletons are still a very common pattern throughout ASP.NET Core. Otherwise, all services would be scoped or transient, and there would be no singletons at all. Admittedly those generally don't offer IDisposable interfaces, but there are certainly other things that can "leak" between requests. In my opinion, this is an acceptable, if non-ideal, case.

bobbyangers:
First of all thank you for acknowledge my PR and taking the time to explain.

I do totally understand that connection pool and bootstrapping is an expensive task that should be performed only once the application is booted up.

Ideally, an **IBucket** should "rent" a connection, and let it go ASAP;  it should not "stick around" in between calls.

The pool should create the connections and manage the health of those connections (which I am sure it does)  and recycle any connection that have "fallen off"; 

That would make the system super resilient and "self-healing".  And ensure that users of the **IBucket** need not worry so much about the lifetime of the object.

My reflex when dealing a disposable object is to use "using";
using var **someObject**  
which as you know is suppose to "freeup" the object ASAP, ie connection are important and expensive resources;  
In this case **using**, it _breaks_, not so great.  

Many examples in the .NET framework, namely in the data access (ie entity framework), the connection pool is managed outside of the connection object hence a limited amount of connection 5-6; can handle literally dozens of instances of connections because connections are short-lived, while the connection pool is long-lived.

That is why, I have to say that I do not agree on with your statement that "this is an acceptable case", simply because I think there is an alternative way, that _would not break the existing semantics_.

brantburnett:
@bobbyangers 

First of all, let me say that I generally completely agree with you about connection pooling as an approach. It is indeed how .NET manages things like SqlConnection within System.Data. They keep a pool of connection objects, anytime you try to connect using the same connection string it just grabs from the pool. And there is an idle timeout which causes cleanup eventually. The current IDisposable approach is a bit confusing, and I haven't been a fan of it for years.

Unfortunately, I don't believe that is currently an option with Couchbase SDKs (though I could be wrong). The Couchbase SDKs are specifically designed to have the same semantics across all SDKs for all languages. The idea is that developers can easily use Couchbase across a variety of languages with a minimal learning curve. This is especially valuable in microservice architectures, where different services may use different languages. While there are clearly variations, core semantics like how to connect to a cluster/bucket and which you can then close that are globally defined. This was implemented in .NET using the closest paradigm, IDisposable.

Someone else may speak up and disagree with me, I'm not an SDK team member just a community contributor. Just giving my two cents as to why it's done this way.

What I'd like to examine a bit is how it's handled within Couchbase.Extensions.DependencyInjection. That is a purely .NET construct, which gives us more leeway. Perhaps we could tweak its internal behavior so at least the IDisposable implementations on IClusterProvider and IBucketProvider aren't as problematic? We'd have to leave them implementing IDisposable for SemVer reasons, but we could make that a no-op and handle shutdown via ICouchbaseLifetimeManager only.

bobbyangers:
The problem I am having, is that if something F**up in on request.... like for some reason using Linq2Couchbase is causing a problem after running a query..  My whole api is rendered inoperable.... until reset.... it's a timebomb.

Any calls to any functions after fails...
[18:37:14 ERR] An unhandled exception has occurred while executing the request. 
System.ObjectDisposedException: The CancellationTokenSource has been disposed.
   at System.Threading.CancellationTokenSource.ThrowObjectDisposedException()
   at System.Threading.CancellationTokenSource.get_Token()
   at Couchbase.Core.Configuration.Server.ConfigHandler.Start(Boolean withPolling)
   at Couchbase.Core.ClusterContext.StartConfigListening()
   at Couchbase.Cluster..ctor(ClusterOptions clusterOptions)
   at Couchbase.Cluster.ConnectAsync(ClusterOptions options)
   at Couchbase.Extensions.DependencyInjection.Internal.BucketProvider.<GetBucketAsync>b__4_0(String name)

bobbyangers:
I am loooking at the constructor for the CouchbaseBucket... everything is lazy loaded....

        internal CouchbaseBucket(string name, ClusterContext context, IScopeFactory scopeFactory, IRetryOrchestrator retryOrchestrator,
            IVBucketKeyMapperFactory vBucketKeyMapperFactory, ILogger<CouchbaseBucket> logger, IRedactor redactor, IBootstrapperFactory bootstrapperFactory,
            IRequestTracer tracer)
            : base(name, context, scopeFactory, retryOrchestrator, logger, redactor, bootstrapperFactory, tracer)
        {
            _vBucketKeyMapperFactory = vBucketKeyMapperFactory ?? throw new ArgumentNullException(nameof(vBucketKeyMapperFactory));

            _viewClientLazy = new Lazy<IViewClient>(() =>
                context.ServiceProvider.GetRequiredService<IViewClient>()
            );
            _viewManagerLazy = new Lazy<IViewIndexManager>(() =>
                new ViewIndexManager(name,
                    context.ServiceProvider.GetRequiredService<IServiceUriProvider>(),
                    context.ServiceProvider.GetRequiredService<CouchbaseHttpClient>(),
                    context.ServiceProvider.GetRequiredService<ILogger<ViewIndexManager>>(),
                    redactor));

            _collectionManagerLazy = new Lazy<ICouchbaseCollectionManager>(() =>
                new CollectionManager(name,
                    context.ServiceProvider.GetRequiredService<IServiceUriProvider>(),
                    context.ServiceProvider.GetRequiredService<CouchbaseHttpClient>(),
                    context.ServiceProvider.GetRequiredService<ILogger<CollectionManager>>(),
                    redactor)
            );
        }
There is even an event to "reset" if the configuration changes
   public override async Task ConfigUpdatedAsync(BucketConfig config)

So making the provider "Scoped", I don't think it would be that much of a deal.   
Yeah, a performance tradeoff, but for a (relatively) low volume; would not make a big difference.

Having this said;  I suggested we have both options;  singleton and scoped through two different extensions methods;  this way the user (me) can choose which one best suit my scenario.

jeffrymorris:
Hi @bobbyangers - 

Thanks for the contribution, but I am going to have to side with @brantburnett on this one. Both ICluster and IBucket implementations hold resources that should be created when the application starts up and released to the OS when the application shuts down. The lifespan of these objects is tied to the application lifespan. This is documented [here](https://docs.couchbase.com/dotnet-sdk/current/howtos/managing-connections.html#connection-lifecycle). 

Thanks,

Jeff

bobbyangers:
@jeffrymorris   Thanks for the input.   I see and I read the reference.

Please tell me, how do you deal with a corrupt bucket ? In a multiple / client web application;  Is the application expected to stop working because of a bad request or misusage in one function/method ?

The fact remains that simply asking the developers not to call a public method, in this case Dispose(); 'cause it'll break the whole system, is not great design. it's plain wrong.   Sharing the cluster I understand;  Sharing the bucket instance, I am not convinced.

jeffrymorris:
@bobbyangers 

>Please tell me, how do you deal with a corrupt bucket ? In a multiple / client web application; Is the application expected to stop working because of a bad request or misusage in one function/method ?

I am not sure I understand what you mean by a "corrupt" bucket; you can Dispose of a bucket whenever you want, you'll just have to a) recreate it the next time you use want to use it b) deal with the performance hit of allocating sockets and doing server negotiation. That being said, once you dispose of a bucket, it cannot be re-used - calling dispose effectively shuts down all sockets and reclaims any resources.

>The fact remains that simply asking the developers not to call a public method, in this case Dispose(); 'cause it'll break the whole system, is not great design. it's plain wrong.

I am not sure I understand this, we never tell developers to not call dispose; however, you should be aware of the ramifications if you do and where you do it.

> Sharing the cluster I understand; Sharing the bucket instance, I am not convinced.

There are both cluster level and bucket level resources that need to be managed; its a server level concept: GCCCP and CCCP.

From reading through the comments, I get the feeling you expect the Couchbase SDK/API to have the same behavior as the ADO.NET API. Simply put they are different and its by design. The type of connection pooling your used doesn't apply to the  Couchbase SDK - there is no static process level pool; each client instance manages its own pool. 

-Jeff

bobbyangers:
@jeffrymorris 
Thanks you so much for your detailed explanations.

I recently have tried to move to version 3.x of the api; and was having many problems.    I am using the Dependency Injection module which is using providers that keep a reference to the **IBucket** that they generate;  however this provider do not track if there "children" have been disposed or not.  So one bad "using", the whole system was "down".  If the providers had a way to better manage the lifetime of the children; it solve all my problems.
  
Which is why I had proposed to make the providers "scoped" instead of "singleton";   I do understand your explanation regarding the resource (socket, etc) management and performance related issues, and that recreating them at each request, might not be the best option.

Would it be possible for the **IBucket** to expose a delegate or event; where anyone referencing it could be notified if the bucket is disposed ?   In this case, the providers could simply "register" with the **IBucket**'s event; and if someone decides to "flush" it; then would simply remove the reference, that way, the next caller (request) would get a brand new **IBucket** ?

brantburnett:
I do kind of like the idea of "rebootstrapping" the IBucket if it is accidentally disposed and then another request calls ICluster.BucketAsync again for that bucket. It kind of makes sense. I just have a few concerns about it.

1. In order to keep performance high, I think it would need to be lock-free on the happy path (bucket isn't disposed and is still available)
2. There are three ways to approach, with pros and cons:
    a. Attach an event for dispose, when received immediately bootstrap again
    b. Attach an event for dispose, but just use it to do cleanup (so GC can collect), wait for next call to BucketAsync to bootstrap
    c. Set a flag on dispose so the next call to BucketAsync sees the flag and triggers new bootstrap
3. Attaching events (via an event sink interface or as .NET events) tends to cause memory leaks if not done carefully, so this would need to be watched for.
4. **Most concerning:** This still wouldn't prevent errors in the application, it would simply cause the application to repair itself. Due to multithreading, it's possible, even likely under load, that some requests would already have the IBucket being disposed and continue to try to use it. Any other in-flight requests would fail when the IBucket was disposed, only subsequent requests would succeed. To make things worse, this likely wouldn't even be noticed until the app reaches production, local dev would probably work fine, unless the user has load testing in their CI pipeline. As it stands today, most kinds of automated testing in the CI pipeline would likely catch the design flaw where Dispose was caused. With this fix in place, it requires parallel requests to trigger failures.
5. Due to #4 and due to performance, this scenario happening would still represent an anti-pattern, so we'd want to log warnings left and right so that developers could find and fix their incorrect Dispose calls

bobbyangers:
Regarding your concerns about performance; I get it;  I value resilience (and stability) over performance;  both are important factor; but different context, the weight of each might be different;

according with @jeffrymorris 
> I am not sure I understand this, we never tell developers to not call dispose; however, you should be aware of the ramifications if you do and where you do it.

> Due to multithreading, it's possible, even likely under load, that some requests would already have the IBucket being disposed and continue to try to use it.

It's not an antipattern to "self-heal";   crashing the whole because of a misuse of a singleton resource is.  This is the case now; no worse than now.   And having all current inflight fail is ok, since the connection is shared.   Dotnet core (web) has a middleware pipeline that takes care of exceptions (if done right);  no need to log "left-right-center"; and they would not go "unnoticed";  

I respectfully disagree that it would make things worse.

brantburnett:
@bobbyangers 

I believe you misinterpreted what was attempting to say. I **do** believe that self-healing is a good approach. I just simply wanted to raise some concerns that I think we need to decide how to address before moving forward with any implementation.

Also, I don't think self-healing is an anti-pattern. It's actually intrinsic to a lot of what the Couchbase SDK tries to do. I believe that calling Dispose on the bucket during a particular HTTP request and then getting the bucket again on other HTTP requests is an antipattern for 99.9% of uses cases. Thus why I feel logging a warning saying "hey, you may be doing this wrong!" is important. It will just help prevent confusing problems for developers down the line.

Note: This comes from personal experience. When I first started using Couchbase back in 2012, I was disposing on every HTTP request and recreating because I didn't realize the proper use pattern. In that case (on an older SDK, clearly), it was causing my application to crash in production every few hours. Even worse, it was happening over Christmas. There went my vacation. Having logs like this would have really saved my bacon.

bobbyangers:
Hi Brant.

Sorry for the delay in response.

I do agree that adding logging would help identify misuse or odd cases.

the end all, is the ensure the namedbucketprovider doesn't blindly assume the bucket it created the first time is still valid at all times.

So either the bucket has a flag object disposed set when dispose qas call on it, or exposes a delegate or a callback, that is calls when disposing, so that provider is made aware that the object is no longer valid, and would need to create a new one.

Of course this should not happen so often.  But when it does...

I would prefer to keep the reference through the depency injection and let the container manage the lifetime.  

Also noting, the IHttpFactory which implements a similar behavior to avoid websocket depletion.  HttpClient is also recommended to be used a singleton and reused, and the lifetime of the client is fully managed by the IHttpClientFactory / and dependency injection; maybe we should look how that is implemented, and to see if we can find some inspiration there.

daschl:
@bobbyangers after careful consideration inside the team we have decided that we're not going to take the pull request for the reasons discussed above in the ticket. But I really want to thank you very much for helping us make the SDK as good as we can and also discuss/scrutinize the architectural decisions.

We hope this doesn't discourage you from further contributions - also if you want to discuss certain larger changes upfront we are happy to discuss them (i.e. in the forums, via email etc.)

Thanks much again!

PR #110, opened 2020-12-09
Possible deadlock

terkill:
I'm registering the buckets with an Autofac container, so I'm using `.GetAwaiter().GetResult()` to get the IBucket instance.
This is working in all my projects except one where the second call to Cluster.BucketAsync hangs. I have been unable to reproduce this in a small project, as I am clueless why it is happining in the one project and not the others.
However, I tracked the problem down to a missing `ConfigureAwait(false)`

### DIFF
--- a/src/Couchbase/DnsClientDnsResolver.cs
+++ b/src/Couchbase/DnsClientDnsResolver.cs
@@ -42,7 +42,7 @@ namespace Couchbase
         public async Task<IPAddress?> GetIpAddressAsync(string hostName,
             CancellationToken cancellationToken = default)
         {
-            var addresses = (IEnumerable<IPAddress>) await _dotNetDnsClient.GetHostAddressesAsync(hostName);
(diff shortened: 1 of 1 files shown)


jeff-longino:
Having recently attempted a port from the 2.x client lib to the 3.x in my app I have been concerned about just this issue.  Can someone on the CB client lib team confirm that the goal is for all awaits within the lib itself to be set to ConfigureAwait(false)?

RiPont:
> Having recently attempted a port from the 2.x client lib to the 3.x in my app I have been concerned about just this issue. Can someone on the CB client lib team confirm that the goal is for all awaits within the lib itself to be set to ConfigureAwait(false)?

Correct.  That is the goal.

jeffrymorris:
@terkill -

Thanks for submitting, this has been merged for 3.1.1.

-Jeff

PR #113, opened 2021-04-13
fix debug format string

smitcham:
The debug string is causing a format exception when
going down this path


smitcham:
The code review site has the PR matched against the 'master' branch instead of the release27 branch, is that an issue, or is there another way to create the review?

smitcham:
Looks like this isn't needed anymore.

PR #115, opened 2021-09-15
Use correct service type name in query context

smitcham:
The N1QL service query was using the nameof operator
which doesn't match in the orphan reporter that is
looking for the value of the property and not the name


smitcham:
I'll reopen from my home computer.  Pushing from work keeps me from being able to sign the agreement.

PR #138, opened 2026-07-15
NCBC-4256: Use OIDC for release S3 publish and smoke-test AWS on publ…

davidkelly:
…ish=false

Motivation
==========
The migrated release workflow (NCBC-4223) authenticated the S3 publish with static AWS keys (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY) that were never provisioned as GitHub secrets. The distribution account (786014483886) already exposes a GitHub OIDC role (SDK_GHA), so the release can federate instead of storing and rotating long-lived keys. Separately, a publish=false smoke run had no way to validate S3 access ahead of a real release.

Modifications
=============
- publish-s3: assume SDK_GHA via aws-actions/configure-aws-credentials OIDC (permissions: id-token: write) instead of static keys; drop both AWS_* secret references. Region us-west-1.
- smoke-summary (publish=false): assume the same role and perform a reversible S3 round-trip under clients/net/_smoke/ -- put with --acl public-read, verify the object is publicly readable, best-effort delete -- exercising OIDC trust + s3:PutObject + s3:PutObjectAcl without publishing a real artifact. Surface the result in the run summary.

publish=false now validates build + sign + pack + AWS access in a single run. The NuGet push credential remains unexercised until a real publish, since nuget.org is immutable and cannot be smoke-tested reversibly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -344,11 +344,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: [validate-inputs, pack]
     if: needs.validate-inputs.outputs.publish == 'true'
+    permissions:
+      id-token: write   # required to mint the GitHub OIDC token for AWS
+      contents: read
     env:
       VERSION: ${{ needs.validate-inputs.outputs.version }}
       MINOR: ${{ needs.validate-inputs.outputs.minor }}
       S3_BUCKET: packages.couchbase.com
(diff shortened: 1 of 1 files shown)


jeffrymorris (approved):
LGTM - maybe reformat the summary and shorten the title.

PR #143, opened 2026-07-20
NCBC-4259 Base Stellar timeout ambiguity on read-only, not idempotent

davidkelly:
Motivation
==========
RFC 77 requires ambiguous-vs-unambiguous timeout
classification to be based on read-only status, not idempotency: an idempotent-but-mutating op (e.g.
GetAndLock, GetAndTouch, MutateIn) is safe to retry yet may still have applied on timeout, so it must
surface as AmbiguousTimeoutException. Stellar keyed the decision on StellarRequest.Idempotent, matching neither the RFC nor the classic SDK (where
IRequest.Idempotent is defined as IsReadOnly).

Modification
============
Add a ReadOnly property to StellarRequest that
defaults to Idempotent (echoing the classic coupling) so existing call sites keep their behavior, and set ReadOnly=false explicitly on the three KV operations that are idempotent yet mutate: GetAndLock,
GetAndTouch, MutateIn. Switch the two timeout
decisions in StellarRetryHandler (token cancellation and DEADLINE_EXCEEDED) from Idempotent to a ReadOnly check via an IsReadOnly helper that falls back to
Idempotent for non-StellarRequest inputs.

Results
=======
18 StellarRetryHandler tests passing, including a new regression test that a timed-out idempotent-but-
mutating op raises AmbiguousTimeoutException and a test of the ReadOnly/Idempotent default coupling.
Retry-strategy behavior is unchanged (Idempotent is untouched).

### DIFF
--- a/src/Couchbase/Stellar/Core/Retry/StellarRequest.cs
+++ b/src/Couchbase/Stellar/Core/Retry/StellarRequest.cs
@@ -33,22 +33,14 @@ public class StellarRequest : IRequest
     public uint Attempts { get; set; }
     public bool Idempotent { get; set; }
 
-    private bool? _readOnly;
-
     /// <summary>
-    /// Whether the operation only reads server state (i.e. never mutates data, expiry, or locks).
-    /// This drives ambiguous-vs-unambiguous timeout classification: a read-only op that times out
-    /// definitively did not mutate state (<see cref="Couchbase.Core.Exceptions.UnambiguousTimeoutException"/>),
-    /// whereas a mutating op might have (<see cref="Couchbase.Core.Exceptions.AmbiguousTimeoutException"/>).
-    /// Mirrors the classic SDK, where <c>IRequest.Idempotent</c> is defined as <c>IsReadOnly</c>
-    /// (see OperationBase). Defaults to <see cref="Idempotent"/> when not set explicitly, as the two
-    /// coincide for most operations; KV operations that are idempotent yet mutate state
-    /// (GetAndLock, GetAndTouch, MutateIn) set this to <c>false</c> explicitly.
+    /// Whether the operation only reads server state. Drives timeout ambiguity classification
+    /// (see <see cref="StellarRetryHandler"/>). Defaults to <see cref="Idempotent"/> when not set.
     /// </summary>
-    public bool ReadOnly
+    public bool? ReadOnly
     {
-        get => _readOnly ?? Idempotent;
-        set => _readOnly = value;
+        get => field ?? Idempotent;
+        set;
     }
 
     public List<RetryReason> RetryReasons { get; set; } = new();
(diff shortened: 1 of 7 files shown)


src/Couchbase/Stellar/Core/Retry/StellarRequest.cs
```diff
@@ -32,6 +32,25 @@ internal StellarRequest(TimeProvider timeProvider)
 
     public uint Attempts { get; set; }
     public bool Idempotent { get; set; }
+
+    private bool? _readOnly;
+
+    /// <summary>
+    /// Whether the operation only reads server state (i.e. never mutates data, expiry, or locks).
```
emilienbev: The code comments are pretty long to explain the rationale of `ReadOnly`, and it's re-explained it on the callers as well.

I'd vote to boil them down to the essential and/or remove the duplicated explanations.

src/Couchbase/Stellar/Core/Retry/StellarRequest.cs
```diff
@@ -32,6 +32,25 @@ internal StellarRequest(TimeProvider timeProvider)
...
+    /// This drives ambiguous-vs-unambiguous timeout classification: a read-only op that times out
+    /// definitively did not mutate state (<see cref="Couchbase.Core.Exceptions.UnambiguousTimeoutException"/>),
+    /// whereas a mutating op might have (<see cref="Couchbase.Core.Exceptions.AmbiguousTimeoutException"/>).
+    /// Mirrors the classic SDK, where <c>IRequest.Idempotent</c> is defined as <c>IsReadOnly</c>
+    /// (see OperationBase). Defaults to <see cref="Idempotent"/> when not set explicitly, as the two
+    /// coincide for most operations; KV operations that are idempotent yet mutate state
+    /// (GetAndLock, GetAndTouch, MutateIn) set this to <c>false</c> explicitly.
+    /// </summary>
+    public bool ReadOnly
```
emilienbev: NIT but we can use the field backed property from the new C# here so we don't to write the backing field!
i.e. remove `private bool? _readOnly` and just do:
```
public bool? ReadOnly
    {
        get => field ?? Idempotent;
        set;
    }
```

emilienbev:
It's not included in this PR but the Query and Analytics requests hardcode `Idempotent == true`. For both Stellar{Query/Analytics}Client it should read it from the options (which your code will read via `StellarRequest.ReadOnly`):
```
var stellarRequest = new StellarRequest
{
    Idempotent = opts.ReadOnly ?? false,
    Token = opts.Token,
    Timeout = opts.TimeOut ?? _stellarCluster.ClusterOptions.QueryTimeout // replaced with AnalyticsTimeout for StellarAnalyticsCLient
};
```

emilienbev (changes requested):

emilienbev (approved):

PR #144, opened 2026-07-20
NCBC-4260: Retry LOCKED documents instead of failing immediately

davidkelly:
Motivation
==========
CNG reports a locked document as a PreconditionFailure "LOCKED" detail block. Stellar mapped that block to an immediate UnambiguousTimeoutException("Document is locked") with no retry, so a transiently locked
document failed the operation outright. Per RFC 77 a locked document is a retryable condition; Java retries via KV_LOCKED and Go via KVLockedRetryReason.

Modification
============
In StellarRetryHandler, change the PreconditionFailure LOCKED case to add the KvLocked retry reason, increment Attempts, and return so the retry loop backs off and retries until the timeout is exceeded, mirroring the existing FailedPrecondition/LOCKED path. Uses return (not break) so it does not fall through into the status switch.

Results
=======
17 StellarRetryHandler tests passing, including a new regression test asserting a LOCKED PreconditionFailure is retried (rather than throwing immediately) and then succeeds.

### DIFF
--- a/src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
+++ b/src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
@@ -189,13 +189,8 @@ internal class StellarRetryHandler : IRetryOrchestrator
                                 throw new CasMismatchException(context);
                             case StellarRetryStrings.PreconditionLocked:
                             {
-                                // CNG reports a locked document as a PreconditionFailure "LOCKED"
-                                // detail block. Per RFC 77 this is a retryable condition (KV_LOCKED):
-                                // the document is only temporarily locked, so retry with backoff until
-                                // the timeout is exceeded rather than failing immediately. Mirrors the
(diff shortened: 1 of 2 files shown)


src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
```diff
@@ -189,8 +189,16 @@ private void HandleException(RpcException protoException, IRequest request, Gene
                                 throw new CasMismatchException(context);
                             case StellarRetryStrings.PreconditionLocked:
                             {
+                                // CNG reports a locked document as a PreconditionFailure "LOCKED"
```
emilienbev: Same nit here as the other PRs, the loong comment is a bit unnecessary. 
We could keep "Retryable condition per RFC 77 (KV_LOCKED)" but the rest is a lot.

emilienbev (approved):

PR #145, opened 2026-07-21
NCBC-4263: fail retryable mid-stream errors as RequestCancelled Motivation

davidkelly:
Motivation
==========
RFC 77 requires a retryable error that occurs mid-stream
(after the first response) to fail as RequestCancelled.
Stellar's query and analytics enumerators instead leaked
mid-stream errors as raw Grpc.Core.RpcException.

Modification
============
Add StellarRetryHandler.ThrowMidStreamException, reusing
HandleException so terminal codes throw their mapped
exception while retryable codes and transient transport
failures become RequestCancelled (carrying the error
context). Route the mid-stream MoveNext loops in
StellarQueryResult and ProtoAnalyticsResult through it.
Search is left as-is: it buffers the whole stream inside
RetryAsync, so retrying its mid-stream errors is safe. That
said, this seems inconsistent so we will explore that in
a separate ticket.

Results
=======
21 StellarRetryHandler tests passing, with new cases for
retryable, transient-transport, and terminal mid-stream
errors. First-response retry behavior is unchanged.

### DIFF
--- a/src/Couchbase/Stellar/Analytics/ProtoAnalyticsResult.cs
+++ b/src/Couchbase/Stellar/Analytics/ProtoAnalyticsResult.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Couchbase.Analytics;
 using Couchbase.Core.IO.Serializers;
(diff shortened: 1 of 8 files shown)


jeffrymorris (approved):
LGTM - regression is fixed.

PR #148, opened 2026-07-22
NCBC-4265: Remove Stellar error handling not specified by RFC 77

davidkelly:
Motivation
==========
The .NET Stellar retry handler maps two gRPC statuses that RFC 77 does not specify, and that CNG does not emit in the way the mappings assume:

- FAILED_PRECONDITION with a "CAS" precondition violation was mapped to CasMismatchException. CNG never returns this; a CAS mismatch arrives as ABORTED with a CAS_MISMATCH error-info block, which the handler already maps correctly.
- RESOURCE_EXHAUSTED was mapped to ValueToolargeException. CNG only uses this status for rate-limit errors on an API the SDK does not call, so the mapping mislabels an unrelated condition.

Removing the CAS mapping also exposed a latent issue: an unrecognized FAILED_PRECONDITION fell through to the status switch and was retried until the request timed out, even though FAILED_PRECONDITION is non-retryable per its gRPC contract (LOCKED is the sole RFC 77 exception).

Modification
============
- Remove the PreconditionCas -> CasMismatchException case (and the now-unused StellarRetryStrings.PreconditionCas constant).
- Remove the ResourceExhausted -> ValueToolargeException case so it falls through to CouchbaseException. Oversized documents are unaffected: they are caught by a client-side pre-flight size check in StellarCollection before the request is sent, not by this mapping.
- Make unmapped FAILED_PRECONDITION terminal on both paths so it is never retried (LOCKED aside): a default in the structured precondition-failure switch, and a terminal throw on the status-switch detail-string path.
- Route the three terminal, non-retryable outcomes (unknown status, and the two unmapped-precondition paths) through a single TerminalException helper so they behave identically: record the gRPC status in the error context and throw CouchbaseException.

Results
=======
Full unit suite passes. Adds regression tests asserting: RESOURCE_EXHAUSTED falls back to CouchbaseException (exact type, not ValueToolargeException); an unrecognized structured precondition type ("CAS") is a terminal CouchbaseException thrown on the first attempt; and an unmapped FAILED_PRECONDITION with no detail block is likewise terminal and not retried. The existing oversized-document ValueToolargeException tests continue to pass, confirming they do not depend on the removed mapping.

### DIFF
--- a/src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
+++ b/src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
@@ -214,7 +214,7 @@ internal class StellarRetryHandler : IRetryOrchestrator
                                     new ValueInvalidException(); //Prone to change as it's unclear what this Precondition failure maps to.
                             default:
                                 // Unknown precondition (e.g. the removed "CAS") — terminal, not retried.
-                                throw TerminalException(context, status, detail);
+                                throw CreateTerminalException(context, status, detail);
                         }
                     }
                 }
@@ -246,7 +246,7 @@ internal class StellarRetryHandler : IRetryOrchestrator
                 if (detail.Contains(StellarRetryStrings.PathValueOutOfRange)) throw new NumberTooBigException();
                 if (detail.Contains(StellarRetryStrings.ValueTooLarge)) throw new ValueToolargeException(detail);
                 // Unmapped FAILED_PRECONDITION is non-retryable (LOCKED aside) — terminal.
-                throw TerminalException(context, status, detail);
+                throw CreateTerminalException(context, status, detail);
             case StatusCode.PermissionDenied:
             case StatusCode.Unauthenticated:
                 throw new AuthenticationFailureException(context);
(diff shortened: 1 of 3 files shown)


src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
```diff
@@ -214,6 +212,9 @@ private void HandleException(RpcException protoException, IRequest request, Gene
                             case StellarRetryStrings.PreconditionValueOutOfRange:
                                 throw
                                     new ValueInvalidException(); //Prone to change as it's unclear what this Precondition failure maps to.
+                            default:
+                                // Unknown precondition (e.g. the removed "CAS") — terminal, not retried.
+                                throw TerminalException(context, status, detail);
```
jeffrymorris: Name hear is confusing; I thought it TerminalException was the class name. I guess it makes sense though because of the `throw` keyword.

jeffrymorris (approved):

src/Couchbase/Stellar/Core/Retry/StellarRetryHandler.cs
davidkelly: Agreed!   Simple enough to name this somewhat better!

PR #150, opened 2026-07-25
NCBC-4269: Implement WaitUntilReady for Stellar

davidkelly:
Motivation
==========
Stellar's cluster and bucket WaitUntilReady threw
FeatureNotAvailableException. RFC 77 defines couchbase2 WaitUntilReady as pinging the standard gRPC health-check RPC until the server reports SERVING, giving the user confidence they can reach a couchbase2 server.

Modification
============
Add the Grpc.HealthCheck package and a Health.HealthClient over the existing GrpcChannel. WaitUntilReadyAsync sends Health.Check with an unset service field through the standard couchbase2 retry rules: SERVING succeeds; a non-SERVING status or UNAVAILABLE/transient error retries; the shrinking per-attempt deadline bounds the operation and surfaces as UnambiguousTimeoutException. No Authorization header is sent and no metrics are emitted, per the RFC. The ServiceTypes and DesiredState options are silently ignored (the health check has no per-service or cluster-state granularity). Bucket delegates to the cluster (identical implementation per the RFC). The lastException diagnostic is deferred to CNG-4 (PENDING in the RFC).

Results
=======
Unit and Fit tests passing, you can see the WaitUntilReady logging in fit so this is working great.

### DIFF
--- a/src/Couchbase/Stellar/StellarCluster.cs
+++ b/src/Couchbase/Stellar/StellarCluster.cs
@@ -77,8 +77,13 @@ internal class StellarCluster : ICluster, IBootstrappable, IClusterExtended
     private readonly SocketsHttpHandler? _socketsHandler;
 
     // Drives the WaitUntilReady timeout budget. Injectable so tests can run the health-check
-    // retry loop on a FakeTimeProvider instead of the wall clock.
+    // retry loop on a FakeTimeProvider instead of the wall clock. The initializer is what the
+    // public ClusterOptions constructor below gets; the internal constructor overwrites it.
     private readonly TimeProvider _timeProvider = TimeProvider.System;
+
+    // The trace span name for WaitUntilReady. Not in OuterRequestSpans: it names no service and is
+    // not an operation span.
+    private const string WaitUntilReadySpanName = "wait_until_ready";
 
     internal Health.HealthClient HealthClient { get; set; }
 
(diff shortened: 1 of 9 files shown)


emilienbev (changes requested):
LGTM logic-wise but I think we can throw away all the comments added to `StellarCluster.cs`.
It doesn't matter to me as much for the test files, but I'd rather we keep the main code on the cleaner side

src/Couchbase/Stellar/StellarCluster.cs
```diff
@@ -353,8 +364,70 @@ public async Task<ISearchResult> SearchQueryAsync(string indexName, ISearchQuery
...
+                var response = await HealthClient.CheckAsync(new HealthCheckRequest(), callOptions);
+
+                if (response.Status != HealthCheckResponse.Types.ServingStatus.Serving)
+                {
+                    throw new RpcException(new Status(StatusCode.Unavailable,
+                        $"couchbase2 server reported serving status {response.Status}"));
+                }
+
+                return WaitUntilReadyResult.Instance;
+            }, request).ConfigureAwait(false);
```
jeffrymorris, line 440: CancellationToken support here?

src/Couchbase/Stellar/StellarCluster.cs
```diff
@@ -353,8 +364,70 @@ public async Task<ISearchResult> SearchQueryAsync(string indexName, ISearchQuery
         return await _searchClient.QueryAsync(indexName, query, options).ConfigureAwait(false);
     }
 
-    public Task WaitUntilReadyAsync(TimeSpan timeout, WaitUntilReadyOptions? options = null)=>
-        throw ThrowHelper.ThrowFeatureNotAvailableException("WaitUntilReady", "Protostellar");
+    public async Task WaitUntilReadyAsync(TimeSpan timeout, WaitUntilReadyOptions? options = null)
+    {
```
jeffrymorris, line 373: I am not sure it's needed for this release, but it seems like WaitUntilReady methods should have overloads for external CancellationToken's, however, the API RFC kind of formed the behaviour in an odd way.

jeffrymorris (commented):
A couple of thought's, nothing terminal.

src/Couchbase/Stellar/StellarCluster.cs
davidkelly: Yea - good point!   Simple enough to add.

jeffrymorris (approved):
LGTM

PR #151, opened 2026-07-27
NCBC-4270: Handle lookupIn specs with no contentAs

emilienbev:
The FIT performer dereferences `LookupInSpec.content_as` unconditionally. That is an optional proto field, so a spec built without it makes the performer throw an NRE before the SDK call is made. 
The test driver then sees an unexpected exception type instead of the `CouchbaseException` or `InvalidArgumentException` it expects.

`SubdocServerErrorsTest.lookupInTooManyCommands` sends 17 get-specs with no content_as, so it fails on both the classic and the couchbase2 runs.

This PR adds null-check `content_as` on the `LookupIn`, `LookupInAllReplicas` and `LookupInAnyReplica` paths, and report no content for a spec that leaves it unset.


davidkelly (approved):
Lets get a commit message in there, and otherwise LGTM

jeffrymorris (approved):
LGTM

PR #152, opened 2026-07-27
NCBC-4271: Fail AllReplicas on unusable server group

emilienbev:
Motivation
----------
RFC-0078 wants DocumentUnretrievable when a SELECTED_SERVER_GROUP read cannot be done. AnyReplica did that. AllReplicas returned an empty stream instead, which looks the same as a doc with no replicas.

Changes
-------
- GetAllReplicas and LookupInAllReplicas now throw DocumentUnretrievable when the group is unset, missing from the config, or holds no copy of the doc.
- LookupInAllReplicas checks before returning the stream, so it fails at the call site like GetAllReplicas.

Needs the matching ZoneAwareUtils change in the FIT driver https://review.couchbase.org/c/transactions-fit-performer/+/249547

### DIFF
--- a/src/Couchbase/KeyValue/CouchbaseCollection.cs
+++ b/src/Couchbase/KeyValue/CouchbaseCollection.cs
@@ -32,6 +32,7 @@ namespace Couchbase.KeyValue
     internal sealed class CouchbaseCollection : ICouchbaseCollection, IBinaryCollection, IInternalCollection
     {
         public const string DefaultCollectionName = "_default";
+        private const string NoPreferredServerGroupMessage = "No preferred Server group was set in the ClusterOptions.";
         private readonly string? _preferredServerGroup;
         private readonly bool _rangeScanSupported;
         private readonly BucketBase _bucket;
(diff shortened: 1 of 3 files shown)


src/Couchbase/KeyValue/CouchbaseCollection.cs
```diff
@@ -830,9 +828,28 @@ public async Task<ILookupInReplicaResult> LookupInAnyReplicaAsync(string id,
...
+            var opts = options?.AsReadOnly() ?? LookupInAllReplicasOptions.DefaultReadOnly;
+
+            // The stream below only runs once enumerated, so an unusable server group is rejected here
+            // to fail at the call site like LookupInAnyReplica does. What it resolves is handed to the
+            // stream so the key mapping and the config walk are not repeated there.
+            ZoneAwareTarget? target = null;
+            if (opts.ReadPreferenceValue == InternalReadPreference.SelectedServerGroup)
+            {
+                target = ResolveZoneAwareTarget(id);
+            }
```
jeffrymorris: If InternalReadPreference is anything other than SelectedServerGroup then the exception won't be thrown until it enumerates later, which I believe is what the RFC is trying resolve. So, looks good in strict mode, but fallback mode still has the pre-PR "looks fine until you enumerate it" behavior for this specific input.

jeffrymorris (changes requested):
Emelien - I think I found an issue, checkout my comment. Also, can you provide unit tests to validate the behaviour?

Thanks,
Jeff

src/Couchbase/KeyValue/CouchbaseCollection.cs
emilienbev: Ah indeed, I'll work on this

jeffrymorris (approved):

davidkelly (approved):

PR #154, opened 2026-07-29
NCBC-4373: Add a nightly cross-platform unit test workflow

davidkelly:
Reproduces the Jenkins nightly: all three unit test projects on seven OS/arch legs, across net8.0, net10.0 and net48.

The PR gate runs only Couchbase.UnitTests, so the
DependencyInjection and OpenTelemetry unit tests (103 tests) have no CI coverage, and macos-latest being arm64 leaves x64 macOS uncovered. TFMs still come from SdkTestTargets rather than being duplicated in YAML; build-and-test.yml is untouched.

The temporary pull_request trigger must be removed before merge - nothing can run this until it is on the default branch.


.github/workflows/nightly-unit-tests.yml
```diff
@@ -0,0 +1,158 @@
...
+      # from the GHA migration onwards; hardcoding them here would set up the same trap for the
+      # next project added. This mirrors the **UnitTest** glob the Jenkins nightly used.
+      #
+      # Enumerated from the solution rather than the filesystem: it keeps the set to projects the
+      # build step above actually built (which --no-build requires), and avoids `find` resolving to
+      # Windows' unrelated find.exe under Git Bash.
+      #
+      # --blame-hang turns an intermittent hung test host into a fast red plus a dump, instead of a
+      # job that sits until the timeout with no diagnostics.
+      - name: Test - all unit test projects
```
emilienbev, line 90: I think we could simplify this a fair bit by dropping the tfm matrix axis and giving each project its own step. `dotnet test` already runs every framework a project targets (so `net48` still gets covered on the Windows part via SdkTestTargets, something I jsut noticed in `build-and-test.yml` actually. The separate `build-and-test-net48.yml` is redundant we can delete it).

From our friend Claude:
```
      - name: Test - Couchbase.UnitTests
        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
        run: >
          dotnet test tests/Couchbase.UnitTests/Couchbase.UnitTests.csproj
          --configuration Release --no-build
          --logger "trx;LogFilePrefix=Couchbase.UnitTests" --results-directory TestResults
          --blame-hang --blame-hang-timeout 5min --blame-hang-dump-type full

      - name: Test - Couchbase.Extensions.DependencyInjection.UnitTests
        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
        run: >
          dotnet test tests/Couchbase.Extensions.DependencyInjection.UnitTests/Couchbase.Extensions.DependencyInjection.UnitTests.csproj
          --configuration Release --no-build
          --logger "trx;LogFilePrefix=Couchbase.Extensions.DependencyInjection.UnitTests" --results-directory TestResults
          --blame-hang --blame-hang-timeout 5min --blame-hang-dump-type full

      - name: Test - Couchbase.Extensions.OpenTelemetry.UnitTests
        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
        run: >
          dotnet test tests/Couchbase.Extensions.OpenTelemetry.UnitTests/Couchbase.Extensions.OpenTelemetry.UnitTests.csproj
          --configuration Release --no-build
          --logger "trx;LogFilePrefix=Couchbase.Extensions.OpenTelemetry.UnitTests" --results-directory TestResults
          --blame-hang --blame-hang-timeout 5min --blame-hang-dump-type full

```

emilienbev:
LGTM but added a comment thinking of ways we could improve/simplify the workflow. 
Found out the `build-and-test-net48.yml` is redundant, net48 is automatically targeted on WIndows in `build-and-test.yml`, so I think the explicit net4.8 TFM here isn't needed as well.

.github/workflows/nightly-unit-tests.yml
davidkelly: oh interesting - ok yea!   I'll make the change and check that

davidkelly: Agreed about the redundancy - was going to deal with that later but now is perfectly fine.   I wanted to avoid contention (the PR gate has more, brings up timing issues, and figured here we needed a "cleaner" signal), hence the tfm matrix and so on.    Here's what Claude thinks:


> Thanks — took both suggestions.
> 
>   **`LogFilePrefix`:** good catch, and worse than it looks. I pulled the artifact from an earlier run: `Couchbase.UnitTests.trx` had
>   `total="2898"`, i.e. one framework's results, not both. `LogFileName` pins a single path so the frameworks overwrite each other. That means
>   every artifact `build-and-test.yml` has uploaded is missing most of its results — filing that separately rather than widening this PR.
> 
>   **`build-and-test-net48.yml`:** agreed, deleted here. Same commit ran 2754 tests either way, so it was pure duplication at ~2m30s per PR.
>   Worth noting the check name `Build and Unit Tests (Windows .NET Framework 4.8)` will disappear from PRs — nothing requires it, the master
>   ruleset has no required status checks (and is currently disabled), so nothing should break.
> 
>   **On the `tfm` axis** — I tried dropping it and hit something I didn't expect: `dotnet test --framework net48` against a project that
>   *doesn't* target net48 exits 0 having run nothing. No error, no warning (verified with and without `--no-build`). So per-framework legs need
>   the MSBuild `TargetFrameworks` query and the `ran == 0` guard, or a leg can go green having tested nothing.
> 
>   The reason I want the axis isn't really the job names — it's contention. Three test hosts on one runner oversubscribe the CPU, and we hit
>   that on this very PR: `SlicedMemoryOwnerTests` went red on net48 in `build-and-test.yml` while the standalone net48 workflow passed on the
>   *same SHA*. Those tests assert `MemoryPool.Rent(32)` returns exactly 32 bytes, which the API doesn't guarantee — .NET Framework's pool hands
>   back 64 under contention, .NET Core doesn't. Separate ticket for the test.
> 
>   **On the hardcoded steps** — this is the one I'm less sure about, so tell me what you think. The reason I went with discovery is that naming
>   projects by hand is how we got here: `build-and-test.yml` names one, and the DependencyInjection + OpenTelemetry tests (103, including the
>   `Reflection.Emit` proxy generators) have been compiled on every leg but never run since the GHA migration. Jenkins globbed for `**UnitTest**`
>   and always caught all three. My worry is the next `*UnitTests` project gets silently skipped the same way. The cost is that ~50 lines of
>   bash is meaningfully harder to read than three declarative steps, which is a fair hit. If you'd rather have the readable version, a middle
>   option is your three steps plus a two-line check that the solution still contains exactly three unit-test projects, so it fails loudly when
>   someone adds a fourth.
>

emilienbev (approved):

PR #156, opened 2026-08-03
NCBC-4251: Add cancellation/timeout to unbounded semaphore and HTTP w…

jeffrymorris:
…aits

Motivation
==========
Cluster.cs:737, AttemptContext.cs:2054 and :2253 called WaitAsync() with no token/timeout, and ClusterVersionProvider.cs:111 called httpClient.GetAsync(uri) with no CancellationToken.

Modifications
=============
ClusterVersionProvider: plumb a CancellationToken through GetVersionAsync -> DownloadConfigAsync -> HttpClient.GetAsync, and add a catch (OperationCanceledException) filter so our own cancellation isn't swallowed by the per-server retry loop.

IClusterVersionProvider is public and may be implemented externally (e.g. via ClusterOptions.AddService), so adding a member to it would be a binary-breaking change. Cancellation support is instead offered via a new IClusterVersionProviderCancellable interface, discovered through a new ClusterVersionProviderExtensions.GetVersionAsync extension method. Implementations that don't support cancellation have the token silently ignored rather than faked.

AttemptContext.InitAtrIfNeeded: bound the ATR-init lock wait so a stuck holder on a distinct code path (SetAtrPendingIfFirstMutation) can't block this waiter indefinitely. Budget is the attempt's own RemainingUntilExpiration, not KeyValueTimeout (unset by default, and unrelated to this lock); on timeout it raises through the existing CheckExpiryAndThrow/ErrorBuilder path used by every other expiry check in this class, with a defensive fallback throw so we can never fall through into the critical section without holding the lock.

Dropped two changes that didn't hold up:
- Cluster.cs's _bootstrapLock: EnsureBootstrapped() has had no production callers since NCBC-2399 (2020); this bounded a wait nothing reaches, and the cluster-lifetime CancellationTokenSource it would have used is never cancelled anywhere in the SDK, only disposed.
- The rollback semaphore timeout in RollbackWithKv: Task.WhenAll still waits on a genuinely stuck permit-holder regardless of whether other waiters time out, so it doesn't prevent the hang it targeted. It also raced _expirationOvertimeMode, which exists specifically so rollback can exceed the normal budget.

Results
=======
154 unit tests pass across netstandard2.0/2.1, net8.0, and net10.0, including new coverage for the extension method's dispatch (real cancellation when supported, silent no-op when not) and for mid-request cancellation via the caller's token.

Change-Id: I5107d3066bd9d3d3d3247e292060ae6d122d9434

### DIFF
--- a/src/Couchbase/Client/Transactions/AttemptContext.cs
+++ b/src/Couchbase/Client/Transactions/AttemptContext.cs
@@ -2250,7 +2250,23 @@ namespace Couchbase.Client.Transactions
 
             // quick check without taking a lock
             if (_atr != null) return;
-            await _attemptLock.WaitAsync().CAF();
+
+            // Bound the wait by the attempt's own remaining budget - not KeyValueTimeout, which is
+            // unset by default and unrelated to this lock - so a stuck holder can't block this waiter
+            // past the transaction's own expiry.
+            var lockTimeout = _overallContext.RemainingUntilExpiration;
+            if (lockTimeout < TimeSpan.Zero) lockTimeout = TimeSpan.Zero;
+            if (!await _attemptLock.WaitAsync(lockTimeout).CAF())
+            {
+                // We waited the entire remaining transaction budget, so we're expired - no need to
+                // re-check via CheckExpiryAndThrow.
+                _expirationOvertimeMode = true;
+                throw CreateError(this, ErrorClass.FailExpiry)
+                    .Cause(new AttemptExpiredException(this,
+                        $"Expired after {lockTimeout} waiting to initialize the ATR"))
+                    .RaiseException(TransactionOperationFailedException.FinalError.TransactionExpired)
+                    .Build();
+            }
             try
             {
                 // TODO: AtrRepository should be built via factory to actually support mocking.
--- a/src/Couchbase/Core/Version/ClusterVersionProvider.cs
+++ b/src/Couchbase/Core/Version/ClusterVersionProvider.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using System.Threading;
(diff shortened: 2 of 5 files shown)


src/Couchbase/Client/Transactions/AttemptContext.cs
```diff
@@ -2250,7 +2250,20 @@ private async Task InitAtrIfNeeded(ICouchbaseCollection collection, string id, I
...
-            await _attemptLock.WaitAsync().CAF();
+
+            // Bound the wait by the attempt's own remaining budget - not KeyValueTimeout, which is
+            // unset by default and unrelated to this lock - so a stuck holder can't block this waiter
+            // past the transaction's own expiry.
+            var lockTimeout = _overallContext.RemainingUntilExpiration;
+            if (lockTimeout < TimeSpan.Zero) lockTimeout = TimeSpan.Zero;
+            if (!await _attemptLock.WaitAsync(lockTimeout).CAF())
+            {
+                CheckExpiryAndThrow(id, DefaultTestHooks.HOOK_ATR_PENDING);
```
davidkelly: We waited the entire transaction time budget, so we don't need to check if we are expired.   Lets just do 

```
                  _expirationOvertimeMode = true;
                  throw CreateError(this, ErrorClass.FailExpiry)
                      .Cause(new AttemptExpiredException(this,
                          $"Expired after {lockTimeout} waiting to initialize the ATR"))
                      .RaiseException(TransactionOperationFailedException.FinalError.TransactionExpired)
                      .Build();```

The throw of TimeoutException below will not be categorized properly so lets not have that fallback at all.

davidkelly (changes requested):
We should not throw an exception that will end up categorized as Other.   Let's just create the exception we know should be thrown when the WaitUntil times out, since that exhausted the entire transaction timeout budget.

davidkelly (approved):
👍

PR #163, opened 2026-08-11
NCBC-4283: Let FIT runs test a branch's performer image

davidkelly:
[NCBC-4283](https://couchbasecloud.atlassian.net/browse/NCBC-4283)

## Problem

`fit-testing-dotnet.yml` hardcoded `performer: 'dotnet-fit-performer:main'`, so every FIT run tested master's SDK.

The SDK under test is baked into the performer image (built by `publish-fit-performer.yml` from a chosen ref), not taken from the calling workflow's checkout. Selecting a branch in the Actions UI therefore changed *which YAML ran* but not *which code was tested* — the UI implies otherwise, so a PR could pass FIT without its changes ever being exercised. The workaround was a throwaway commit editing the hardcoded tag on the branch.

## Change

Adds an optional `performer_tag` dispatch input, defaulting to the branch the workflow is run from.

Deriving that tag needs shell — `dk/ncbc-4283` is not a valid Docker tag — and a `uses:` job cannot also have `steps`, so a small `performer-tag` job resolves it and passes it via job outputs. It mirrors the two rules the publish side already applies:

- `sdk-docker-build-action` renames `master` → `main`
- `docker/metadata-action` collapses runs of `[^a-zA-Z0-9._-]` to `-`

Verified against the tags actually published to ghcr.io today:

| `GITHUB_REF_NAME` | performer tag |
| --- | --- |
| `master` (nightly) | `main` |
| `dk/ncbc-4300` | `dk-ncbc-4300` |
| `dk/pe-ramp-bootstrap` | `dk-pe-ramp-bootstrap` |
| `NCBC-4271` | `NCBC-4271` |
| `release/3.9.x` | `release-3.9.x` |

The resolved image is echoed into the job log, so a run states what it tested.

The input is passed via `env:` rather than interpolated into the script body, since `${{ }}` is substituted before bash parses it. The tag job gets `permissions: {}` — it has no need for the OIDC token the `fit` job requires.

## Nightly is unchanged

`schedule` passes no inputs and `GITHUB_REF_NAME` is `master`, which resolves to `main` exactly as before.

## Usage

Build the image first, or the run fails to pull it:

1. **Publish FIT Performer Image** with `ref=<your-branch>`
2. **Run FIT-CLI** from `<your-branch>` — no `performer_tag` needed

Pass `performer_tag: main` to test master's SDK from a branch.

[NCBC-4283]: https://couchbasecloud.atlassian.net/browse/NCBC-4283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/fit-testing-dotnet.yml
+++ b/.github/workflows/fit-testing-dotnet.yml
@@ -29,6 +29,14 @@ on:
         required: false
         type: string
         default: ci
+      performer_tag:
+        description: >-
+          Tag of the ghcr.io/couchbase/dotnet-fit-performer image to test. Defaults to the branch
+          this workflow is run from. The SDK under test lives inside that image, so build it first
+          via "Publish FIT Performer Image" with ref=<your-branch> — otherwise the run fails to
+          pull the image. Use "main" to test master's SDK from a branch.
+        required: false
+        type: string
 
 permissions:
   id-token: write   # required for GitHub OIDC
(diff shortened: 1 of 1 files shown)


emilienbev:
We had a "Run FIT-CLI (Shared)" (`fit-testing-dotnet.yml` called it, passing the hardcoded `dotnet-fit-performer:main`) which was the base workflow where we could pass a custom performer image, not sure what happened to that one. 
Edit: Ah found it, it was removed with #157 .

LGTM 👍  after Copilot's comment on the `needs` expression if it's valid.

davidkelly:
Thanks both.

@Copilot's `needs` point is a false positive — hyphens are fine in `needs.<job-id>` dot notation. Two disproofs:

1. This exact commit (`3fb3a7864`) already ran green: [run 31523670346](https://github.com/couchbase/couchbase-net-client/actions/runs/31523670346), where the `fit` job received `performer: dotnet-fit-performer:dk-ncbc-4283` — which can only come from `needs.performer-tag.outputs.tag`. An invalid expression fails evaluation rather than resolving correctly.
2. `apidocs.yml` and `release.yml` already use `needs.validate-inputs.outputs.*` in production.

@emilienbev re the old `fit-testing-shared.yml`: correct that #157 removed it, but it wouldn't have helped — it took a `performer_image_name` input, and `fit-testing-dotnet.yml` passed the hardcoded `'dotnet-fit-performer:main'` into it. The shared fit-cli workflow likewise exposes `performer` as a required input today; the gap was always that the caller never surfaced it.

Also smoke-tested end to end: built `dotnet-fit-performer:dk-ncbc-4283` from this branch, ran `op-onprem-func-sanity` from the branch with no `performer_tag`, and confirmed via the ghcr config blob that the image the AWS host pulled and ran carries `org.opencontainers.image.revision=3fb3a7864` (vs `:main` = `930cad057`). Result: PASS.

PR #167, opened 2026-08-19
NCBC-4287: Frame the collection ID from the connection NCBC-4285: Resolve the collection ID in GetPrimary

davidkelly:
## Summary

KV operations could address the wrong collection.

Whether a request carries a collection-ID prefix on the front of the document key is decided by what the *connection* negotiated with the server. The SDK decided it from its own state instead, so the two could disagree — and when they did, the server read part of the document key as the collection ID, or read the prefix as part of the key. Either way the operation silently hits the wrong collection, and neither side can detect it: both frames are structurally valid and nothing in the packet distinguishes them.

Underneath that, nothing in the code *forced* an operation to resolve a collection ID before sending it. There were 23 hand-copied guards doing it by convention, and an operation that skipped them compiled fine — which is how `GetPrimary` went four years without one.

This PR keys the decision on the connection, funnels every KV operation through a single place that cannot be skipped, and fixes the related defects found on the way — including two that meant a failed connection handshake could never be handled at all.

**2983 unit tests pass on net8.0 and net10.0**; the solution builds clean in Release. 51 tests are new, and every new assertion was checked by deliberately breaking the code to confirm the test catches it (table near the bottom).

Two tickets: **NCBC-4285** is the narrow defect, **NCBC-4287** is why it was possible.

---

## What was wrong, and what changed

Each item stands alone — no ticket needed to follow it.

### 1. The prefix decision came from client state, not the connection

`OperationBase.WriteKey` wrote a collection-ID prefix whenever it happened to be holding a CID (`Cid.HasValue`). Whether the server *parses* one is fixed by what that connection negotiated in HELO. The two disagree in both directions, and both are silent:

- prefix written on a connection that never negotiated collections → the server treats the prefix byte as part of the document ID, computes the vbucket without it, and the document becomes unreachable
- no prefix on a connection that did → the server reads the first byte of the document key as the collection ID

`WriteKey` now takes the connection's negotiated state, which `SendAsync` already holds and reads six lines below for `Json` and `SnappyCompression`.

### 2. Nothing forced an operation to resolve a collection ID

23 copies of the same guard, 19 copies of the identity stamp, and no compiler help. New `PrepareAsync(op, options)` resolves the CID, stamps `BucketName`/`SName`/`CName`/`Cid`, configures the operation and returns the rented token source — **the caller needs that return value in order to dispatch**, so an operation that skips preparation does not compile.

`ThrowIfBootStrapFailed` deliberately stays at the top of each method: folding it in would move an early exit to after the request span is opened, and it was never the defect.

### 3. `GetPrimary` never resolved its collection ID — NCBC-4285

Reached via `GetAllReplicasAsync`. It was a near-identical copy of `GetReplica` twelve lines below, which *did* have the guard — added five weeks later by NCBC-2881, which missed this one. Fixed as a consequence of item 2 rather than by its own patch.

Rarely surfaced because `PopulateCidAsync` short-circuits once a CID is cached, so only a `GetAllReplicasAsync` that was the first such call on a collection instance was exposed. It becomes a reliable wrong answer once a bucket's lifetime collection count passes ~120, where CIDs stop fitting in the single leb128 byte an ASCII key leaves room for.

### 4. A failed HELO was swallowed and the connection kept

The response status was discarded, `GetValue()` returned null on anything but success, and the caller quietly assigned `ServerFeatureSet.Empty` — leaving a connection in the pool with no negotiated features while operations were still framed as though it had them. Fault injection against 3.9.0 showed this produces exactly the NUL-byte-prefixed document IDs reported in the NCBC-4146 support cases. HELO failure now fails the connection.

### 5. …and a failed HELO could never reach its own error handling

In `ClusterNode.ExecuteOp`, success returns early and every other status fell through to `ErrorMap.TryGetGetErrorCode(...)`. The error map is fetched *after* HELO, so it is null, and every connection-initialisation failure became a `NullReferenceException` before any caller saw the status. Item 4 was unreachable until this was null-guarded.

### 6. A missing collection ID went on the wire

New `RequiresCollectionId` on `OperationBase`, defaulting to `true` — the safe answer, so an operation added without a thought about framing fails loudly rather than silently addressing the wrong collection. The 15 connection- and cluster-level operations whose key is not a document ID override it to `false`. A null CID on a collections-negotiated connection now throws instead of being sent.

### 7. Key length ignored the prefix

`Key`'s setter validated the document ID alone against the 250-byte limit, but the key field carries the collection ID too. A 250-byte key therefore failed as a bare `ArgumentOutOfRangeException` from inside `OperationHeader.KeyLength`, deep in the write path. Now validated at framing time with a message that says what happened. Java validates the same way, prefix included.

### 8. Buckets without collections left the CID null

A memcached bucket has only the default collection, but the connection may still have negotiated collections — in which case the server reads a CID from every key, and the right value is 0. Leaving it null is why **every KV operation against a memcached bucket failed with `CollectionNotFound` after burning the full `KvTimeout`**. Now resolves to 0.

### 9. A named collection on a cluster without collections was sent anyway

It cannot be satisfied, so it now throws `FeatureNotAvailableException` at the API boundary instead of failing with `CollectionNotFound` after the full timeout. Matches the JVM's `BaseKeyValueRequest.encodedExternalKeyWithCollection`.

### 10. `_default._default` cost a `GET_CID` round trip

The default scope and collection are defined to be collection ID 0. Now resolved directly, saving a round trip on the first operation against most collections.

### 11. Cluster-wide feature flags were last-write-wins

`ClusterContext.SupportsPreserveTtl` and `SupportsBinaryXattr` describe the cluster but were assigned from whichever node initialised last, at **seven separate sites each overwriting the one before** — so in a mixed-version cluster the answer depended on ordering. Now recomputed as the conjunction over every node that has negotiated, at the membership choke points (`AddNode`/`RemoveNode`) and when a node finishes negotiating, since a node is often added before its features are known. Nodes that have not negotiated are skipped rather than counted as unsupporting, so the flags do not flap to false while a node comes up.

They stay fields rather than computed properties because they are read on every mutation; turning a field read into an enumeration of the node list is not a trade worth making there.

### 12. Transactions used collections support as a proxy for PreserveTtl support

`DocumentRepository.cs` passed `Bucket.SupportsCollections` as the `PreserveTtl` flag, with a comment calling it *"a proxy for supporting TTLs"*. Two unrelated capabilities. Item 11 turns that from flaky into reliably wrong during a rolling upgrade — transactions would request a feature the cluster then refuses. Five call sites now read the real capability, via a helper that falls back to the old proxy for Protostellar, which has no `ClusterContext` to ask.

### 13. `GetCidLazyNoRetry` retried

Both lazies were built with `retryIfFailure: true`, so the no-retry one retried. Currently only reachable via `PopulateCidAsync(retryIfFailure: false, forceUpdate: false)`, which nothing calls — the one non-retry caller passes `forceUpdate: true` and takes a branch that bypasses the lazies. A trap for the next caller rather than a live bug.

### 14. A successful `GET_CID` with an empty body produced a null CID

`GetValueAsUint()` returns null for a Success response whose body is empty or unparseable, and that null was assigned to `Cid` and carried to the wire. It now fails at the lookup, where the cause is still known. The *deliberate* null from the older-server fallback path is untouched — that one means "this server has no collections", which is a different answer.

### 15. `ClusterContext.SupportsCollections` was dead

A public settable property that nothing assigned and nothing read — permanently false while reading as authoritative. Two tests were setting it in object initialisers, which rather makes the case. Removed.

### 16. `ExecuteLookupIn` resolved the CID unconditionally

Not behind `RequiresCid()`, so subdoc lookups forced a `GET_CID` round trip on buckets that do not support collections — plausibly one contributor to item 8.

### 17. `ScanAsync`'s guard was load-bearing

`PartitionScan` builds its own operations and reads `_collection.Cid` as a plain property, so it cannot go through the funnel. Removing `ScanAsync`'s guard broke range scans; caught because 23 removed guards produced only 19 `PrepareAsync` sites. It now calls the extracted `EnsureCollectionIdAsync`, which `PrepareAsync` also uses, so there is one implementation of the resolve.

---

## Behaviour changes to accept consciously

- **Item 11 is a tightening.** In a mixed-version cluster where some nodes lack `PreserveTtl`, callers who currently get it by luck of initialisation order now get `FeatureNotAvailableException`. That is the point; the alternative is sending the flag to a node that does not implement it.
- **Item 9 moves an error earlier** — from a timeout to an immediate throw at the API boundary.
- **Item 16 removes a round trip** on non-collections buckets.
- **The internal `get_cid` span now nests inside the operation's span** rather than preceding it, because the resolve happens inside `PrepareAsync`. Avoiding this would mean resolving before the span opens, i.e. back to a per-method call, which is the property this change exists to remove.
- **`Observe` keeps the old client-side rule.** It frames its key inside the body with an explicit length and is dead code pending removal, so it is passed `Cid.HasValue` to reproduce exactly what it did before.

## How it was verified

Every new assertion was checked by breaking the code and confirming the test fails — not by the test passing.

| Deliberate break | Result |
| --- | --- |
| `PrepareAsync` stops stamping `Cid` | 22 failures, *"dispatched Delete for key 'thekey' with Cid null"* |
| `PrepareAsync` stops resolving | 22 failures, *"never fetched a collection ID"* |
| one entry dropped from the coverage table | *"Add to the collection ID coverage table: DecrementAsync"* |
| resolve removed from `ScanAsync` | exactly 1 failure, `ScanAsync` |
| `WriteKey` reverted to `Cid.HasValue` | both framing-polarity tests fail |
| `SaslStep`'s opt-out removed | exemption-set test fails |
| key-length check removed | max-length-key test fails |
| default-collection fallback removed | memcached test fails |
| HELO status check reverted | HELO test fails |
| error-map null guard removed | HELO test fails with the original NRE |
| boundary throw removed | named-collection test fails |
| default-collection shortcut removed | no-lookup test fails |
| cluster flags reverted to last-write-wins | 2 of its 5 tests fail |
| transactions proxy restored | 2 of its 4 tests fail |
| `GetCidLazyNoRetry` reverted to retrying | no-retry test fails |
| empty-body CID check removed | empty-body test fails |

Two of those tests initially passed while testing nothing: the cluster-feature tests agreed with last-write-wins until the node ordering was flipped so the permissive node is added last. Caught by the mutation, not by review.

**The exemption list in item 6 is pinned by a reflection test** over every `OperationBase` subclass rather than maintained by hand, because drafting it by hand had already produced a bug — `SaslStart` and `SaslList` were exempt but `SaslStep` was not, and since HELO completes before authentication that would have thrown on every non-TLS connection. Only the opt-outs are pinned: `true` is safe, so a new document operation needs no test change; `false` is the answer that can corrupt data.

Not verifiable locally: **net48**, which is Windows-only in `SdkTestTargets` and has no macOS reference assemblies. CI's Windows leg is the check.

## Test fixtures corrected rather than code

- `CouchbaseCollectionTests` and `CouchbaseCollectionZoneAwareTests` built collections with `Mock.Of<IScope>()`, which reports `IsDefaultScope` as false — so every fake collection looked like a *named* collection on a bucket without collections, and item 9's throw fired. They now describe the default scope, which is what they always meant.
- Tests that build packets against a mocked `IConnection` now supply `ServerFeatureSet.Empty`, since the framing path reads it where the old code short-circuited past it. Collections off means their expected bytes are unchanged.

## Deliberately not covered

- **`PartitionScan` stays outside the funnel** (item 17). Its `RangeScanCancel` writes no key, so routing it through `PrepareAsync` would change that frame.
- **`MutateIn`'s `OptionalFlags` still reads cluster-scoped `SubdocBinaryXattr`** rather than the connection's. That is the one remaining place where a *frame* is decided from cluster state, and it needs the same plumbing `WriteKey` just got.
- **The dead `Observe` operation** is left in place.

### DIFF
--- a/src/Couchbase/Client/Transactions/DataAccess/DocumentRepository.cs
+++ b/src/Couchbase/Client/Transactions/DataAccess/DocumentRepository.cs
@@ -51,8 +51,8 @@ namespace Couchbase.Client.Transactions.DataAccess
                 .AccessDeleted(true)
                 .CreateAsDeleted(true);
 
-            // we always preserve TTLs - using SupportsCollections as a proxy for supporting TTLs
-            opts.PreserveTtl(collection.Scope.Bucket.SupportsCollections);
+            // we always preserve TTLs where the cluster can
+            opts.PreserveTtl(collection.SupportsPreserveTtl());
 
             // we set the flags when staging the insert (though, we do it again when we unstage)
             if (SupportsReplaceBodyWithXattr(collection))
@@ -363,7 +363,7 @@ namespace Couchbase.Client.Transactions.DataAccess
             new MutateInOptions().Defaults(_durability, _keyValueTimeout)
                 .Transcoder(Transactions.MetadataTranscoder)
                 .StoreSemantics(storeSemantics)
-                .PreserveTtl(collection.Scope.Bucket.SupportsCollections);
+                .PreserveTtl(collection.SupportsPreserveTtl());
 
         private List<MutateInSpec> CreateMutationSpecs(IAtrRepository atr, string opType, IContentAsWrapper content, string opId, DocumentMetadata? dm = null, DateTimeOffset? expiry = null)
         {
--- a/src/Couchbase/Core/ClusterContext.cs
+++ b/src/Couchbase/Core/ClusterContext.cs
@@ -107,8 +107,6 @@ namespace Couchbase.Core
 
         public ICluster Cluster { get; }
 
-        public bool SupportsCollections { get; set; }
-
         public bool SupportsGlobalConfig { get; private set; }
 
         public bool SupportsPreserveTtl { get; internal set; }
@@ -310,7 +308,45 @@ namespace Couchbase.Core
                 _logger.LogDebug(
                     "Added node {endPoint} on {bucket} to {nodes}",
                     _redactor.SystemData(node.EndPoint), node.Owner?.Name, Nodes.Count);
+
+                RecomputeClusterFeatureSupport();
             }
+        }
+
+        /// <summary>
+        /// Recomputes the cluster-wide feature flags from every node currently managed.
+        /// </summary>
+        /// <remarks>
+        /// SupportsPreserveTtl and SupportsBinaryXattr describe the cluster, but they used to be
+        /// assigned from whichever single node was initialized last - eight separate sites, each
+        /// overwriting the one before, so the answer depended on ordering. A feature is only safe to
+        /// use if every node offers it, so this takes the conjunction. Nodes that have not negotiated
+        /// yet are skipped rather than counted as unsupporting, so the flags do not flap to false
+        /// while a node is still coming up.
+        ///
+        /// This stays a field rather than a computed property because it is read on every mutation.
+        /// </remarks>
+        internal void RecomputeClusterFeatureSupport()
+        {
+            var seenAnyNode = false;
+            var preserveTtl = true;
+            var binaryXattr = true;
+
+            foreach (var node in Nodes)
+            {
+                var features = node.ServerFeatures;
+                if (features is null)
+                {
+                    continue;
+                }
+
+                seenAnyNode = true;
+                preserveTtl &= features.PreserveTtl;
+                binaryXattr &= features.SubdocBinaryXattr;
+            }
+
+            SupportsPreserveTtl = seenAnyNode && preserveTtl;
+            SupportsBinaryXattr = seenAnyNode && binaryXattr;
         }
 
         public bool RemoveNode(IClusterNode removedNode)
@@ -322,6 +358,9 @@ namespace Couchbase.Core
                 _logger.LogDebug(
                     "Removed node {endPoint} from {nodes}", _redactor.SystemData(removedNode.EndPoint), Nodes);
                 removedNode.Dispose();
+
+                //A node leaving can restore a feature the whole cluster now offers.
+                RecomputeClusterFeatureSupport();
                 return true;
             }
             return false;
@@ -493,8 +532,6 @@ namespace Couchbase.Core
                                 node.NodesAdapter = nodeAdapter;
                                 if (node.ServerFeatures != null)
                                 {
-                                    SupportsPreserveTtl = node.ServerFeatures.PreserveTtl;
-                                    SupportsBinaryXattr = node.ServerFeatures.SubdocBinaryXattr;
                                 }
 
                                 AddNode(node);
@@ -511,8 +548,6 @@ namespace Couchbase.Core
 
                                 if (newNode.ServerFeatures != null)
                                 {
-                                    SupportsPreserveTtl = newNode.ServerFeatures.PreserveTtl;
-                                    SupportsBinaryXattr = newNode.ServerFeatures.SubdocBinaryXattr;
                                 }
 
                                 AddNode(newNode);
@@ -825,8 +860,6 @@ namespace Couchbase.Core
                             {
                                 clusterNode.Owner = bucket;
                                 await clusterNode.SelectBucketAsync(bucket.Name, CancellationToken).ConfigureAwait(false);
-                                SupportsPreserveTtl = clusterNode.ServerFeatures.PreserveTtl;
-                                SupportsBinaryXattr = clusterNode.ServerFeatures.SubdocBinaryXattr;
                             }
 
                             continue;
@@ -840,8 +873,6 @@ namespace Couchbase.Core
                             {
                                 clusterNode.Owner = bucket;
                                 await clusterNode.SelectBucketAsync(bucket.Name, CancellationToken).ConfigureAwait(false);
-                                SupportsPreserveTtl = clusterNode.ServerFeatures.PreserveTtl;
-                                SupportsBinaryXattr = clusterNode.ServerFeatures.SubdocBinaryXattr;
                             }
                         }
 
@@ -866,8 +897,6 @@ namespace Couchbase.Core
                         {
                             clusterNode.Owner = bucket;
                             await clusterNode.SelectBucketAsync(bucket.Name, CancellationToken).ConfigureAwait(false);
-                            SupportsPreserveTtl = clusterNode.ServerFeatures.PreserveTtl;
-                            SupportsBinaryXattr = clusterNode.ServerFeatures.SubdocBinaryXattr;
                         }
                     }
                     else
@@ -885,8 +914,6 @@ namespace Couchbase.Core
                         {
                             clusterNode.Owner = bucket;
                             await clusterNode.SelectBucketAsync(bucket.Name, CancellationToken).ConfigureAwait(false);
-                            SupportsPreserveTtl = clusterNode.ServerFeatures.PreserveTtl;
-                            SupportsBinaryXattr = clusterNode.ServerFeatures.SubdocBinaryXattr;
                         }
 
                         AddNode(clusterNode);
@@ -989,10 +1016,6 @@ namespace Couchbase.Core
 
                             if (newNode.ServerFeatures != null)
                             {
-                                SupportsPreserveTtl =
-                                    newNode.ServerFeatures.PreserveTtl;
-                                SupportsBinaryXattr = newNode.ServerFeatures
-                                    .SubdocBinaryXattr;
                             }
 
                             AddNode(newNode);
--- a/src/Couchbase/Core/ClusterNode.cs
+++ b/src/Couchbase/Core/ClusterNode.cs
@@ -367,7 +367,21 @@ namespace Couchbase.Core
 
             return await ExecuteInternalOperationAsync(connection, heloOp,
                 ExecuteOp,
-                static (_, op) => op.GetValue(),
+                static (status, op) =>
+                {
+                    //A failed HELO used to be swallowed: the status was discarded here, GetValue()
+                    //returned null on anything but success, and the caller quietly assigned
+                    //ServerFeatureSet.Empty. The connection then stayed in the pool with no
+                    //negotiated features while still being framed as though it had them, which is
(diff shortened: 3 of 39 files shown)


src/Couchbase/Client/Transactions/DataAccess/DocumentRepository.cs
```diff
@@ -363,7 +363,7 @@ private MutateInOptions GetMutateInOptions(StoreSemantics storeSemantics, ICouch
             new MutateInOptions().Defaults(_durability, _keyValueTimeout)
                 .Transcoder(Transactions.MetadataTranscoder)
                 .StoreSemantics(storeSemantics)
-                .PreserveTtl(collection.Scope.Bucket.SupportsCollections);
+                .PreserveTtl(collection.SupportsPreserveTtl());
```
jeffrymorris, line 366: This seems like a different issue/bug than NCBC-4285?

src/Couchbase/Core/ClusterContext.cs
```diff
@@ -310,9 +308,47 @@ public void AddNode(IClusterNode node)
...
+                }
+
+                seenAnyNode = true;
+                preserveTtl &= features.PreserveTtl;
+                binaryXattr &= features.SubdocBinaryXattr;
+            }
+
+            SupportsPreserveTtl = seenAnyNode && preserveTtl;
+            SupportsBinaryXattr = seenAnyNode && binaryXattr;
+        }
```
jeffrymorris, line 350: What if we have a mixed node cluster, would that cause an issue?

src/Couchbase/Core/ClusterNode.cs
```diff
@@ -367,7 +367,21 @@ private async Task<ServerFeatures[]> Hello(IConnection connection, IRequestSpan
...
+                {
+                    //A failed HELO used to be swallowed: the status was discarded here, GetValue()
+                    //returned null on anything but success, and the caller quietly assigned
+                    //ServerFeatureSet.Empty. The connection then stayed in the pool with no
+                    //negotiated features while still being framed as though it had them, which is
+                    //how a rejected HELO produces corrupted document keys. There is no usable
+                    //connection without a HELO, so fail it.
+                    if (status != ResponseStatus.Success)
+                    {
+                        ThrowHelper.ThrowHelloFailedException(status);
```
jeffrymorris, line 380: Behaviour change (for the better, but still...)?

jeffrymorris (commented):
Just a few questions to answer.

src/Couchbase/Client/Transactions/DataAccess/DocumentRepository.cs
davidkelly: This was using SupportsCollections (from the bucket caps in the cluster map) as a proxy for supporting preserving ttls, which comes from HELO per connection.   That seemed logical since the 2 both arrived in 7.0 servers, but was always a bit of a hack.   It isn't completely clear to me if these 2 can diverge anyways, but this is "more correct".   One could easily argue it is a separate, tiny, PR if you would like.

src/Couchbase/Core/ClusterContext.cs
davidkelly: This is really for that case.   The flags above are meant to be cluster-wide capabilities.   Before, they were set according to whatever the last connection supported.   This fixes that.   

Before in a mixed cluster, you may think you do have preserveTtl but your specific connection doesn't.   You'd get a FeatureNotAvailableException when you wouldn't expect it.

src/Couchbase/Core/ClusterNode.cs
davidkelly: A failed HELO (say, the server returned an error) used to be interpreted as a connection with no server features.   We'd happily use it, and write the CID before the key of the document but send it to a server that had not agreed to collections support, which then interprets the CID and key together as the key (in a vbucket chosen by hashing the key without the CID).   That's a way to end up with the NCBC-4146 issue.

davidkelly:
Closing in favour of four smaller PRs. Same work, same tests, reviewed content unchanged — split so each piece can be read on its own.

| PR | What | Size | Dependencies |
| --- | --- | --- | --- |
| #168 | Cluster feature support derived from every node, and the transactions PreserveTtl capability | 9 files, +301/−32 | none |
| #169 | A failed HELO fails the connection, and can reach its own error handling | 3 files, +66/−2 | none |
| #170 | The `PrepareAsync` funnel, collection ID resolution, and NCBC-4285 | 5 files, +613/−300 | none |
| #171 | Collection-ID framing keyed on the connection | 25 files, +391/−32 | stacked on #170; also needs #169 |

#168, #169 and #170 are independent and can be reviewed in parallel and merged in any order. #171 is stacked on #170's branch rather than trusting merge order, because without the funnel `GetPrimary` resolves no collection ID and #171's throw fires on every `GetAllReplicasAsync` against a collections cluster — and CI would not catch that, since the framing tests use mocked connections with collections off and the coverage table that would catch it lives in #170.

**The split is verified, not assumed.** Merging all four back together reproduces this branch: the union builds clean and runs 2983 tests, the same count as here. The only differences are three files changed deliberately — `ThrowHelper.cs` (helper placement only; all four helpers byte-identical in signature and body), 20 lines of now-dead factory removed from the CID tests, and `PreserveTtlCapabilityTests` rewritten with its own fixture so #168 doesn't secretly depend on #170. Each branch also builds and passes standalone: 2940 / 2932 / 2962 / 2973.

@jeffrymorris — apologies for the churn, and thanks for the review here. Your three questions and the replies are on this PR rather than the new ones, so for the record:

* the transactions PreserveTtl change is in **#168**, and you were right that it is a separate issue — it is a standalone correctness fix, bundled there because it is the same category error in the same area, not because it is unsafe to split
* the mixed-cluster question is **#168** as well: those flags are meant to be cluster-wide capabilities and were being set from whatever the last connection supported
* the HELO behaviour change is **#169**

Nothing in the code changed in response to those replies, so re-reviewing the same content is not needed — #168 and #169 are where those conversations continue if anything is still open.

PR #175, opened 2026-08-31
NCBC-4291: Fix cross-Cluster lost-cleanup interference from a static field

jeffrymorris:
LostTransactionManager.CollectionsToClean was `static`, shared by every LostTransactionManager in the process, even though exactly one is created per Cluster (see Transactions.cs). Two Cluster instances registering cleanup for the same keyspace would GetOrAdd-reuse a single, shared PerCollectionCleaner. Disposing either Cluster's Transactions instance (e.g. a short-lived connection opened just for one operation) then tore down and removed that shared cleaner's client record - silently killing lost cleanup for the *other*, still-active Cluster with no error anywhere.

This was found via FIT: LostCleanupForSDKIntegrationTest and LostTxnsCleanupTest intermittently fail because a forced-expired ATR entry is never picked up, even though the cleaner's own logs show it completing full, clean scan passes throughout the test's wait window - consistent with the cleaner responsible for that keyspace having been disposed by an unrelated connection's teardown partway through the test class (which does open extra per-test connections with cleanup enabled alongside the shared one, matching the "2 simultaneously connected Cluster instances" diagnostic warning seen in the same test runs).

Fix: make CollectionsToClean an instance field, so each Cluster's LostTransactionManager owns and disposes only its own cleaners. Added DisposingOneManager_DoesNotAffectAnotherManagers_RegisteredCollections, verified to fail against the old `static` field and pass with this fix.

Change-Id: I4a302fdc3388aebfdf431e28e6187942dadd7320

### DIFF
--- a/src/Couchbase/Client/Transactions/Cleanup/LostTransactions/LostTransactionManager.cs
+++ b/src/Couchbase/Client/Transactions/Cleanup/LostTransactions/LostTransactionManager.cs
@@ -29,8 +29,17 @@ namespace Couchbase.Client.Transactions.Cleanup.LostTransactions
         // generating function for the value under contention when it doesn't actually end up doing an insert
         // because someone else got there first.   The Lazy<T> insures the actual initialization of the lazy is
         // only done once.   You can see tests fail when this isn't the case.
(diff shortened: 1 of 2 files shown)


davidkelly (approved):
👍 The copilot comment above seems worth handling though.

PR #181, opened 2026-09-04
NCBC-4296: Honour the configured redaction level in error contexts

davidkelly:
`ClusterOptions.RedactionLevel` redacted the log stream but not exception error contexts, which serialized every field raw — so a customer who turned redaction on for compliance got logs that *looked* clean while document keys, statements and query parameters went out untouched. This applies the redactor at every error-context construction site, matching Java's classification, and reuses the user/meta/system classification the SDK's log paths already use. Redaction is off by default so nothing changes unless a customer opted in; for those who did, `ex.Context.DocumentKey` now returns `<ud>the-key</ud>`, which needs a release note.

Targets **`3.10`**, and has been rebased onto it twice — see *Rebased onto 3.10* below.

## What's classified as what

| Classification | Fields |
|---|---|
| user data (`<ud>`, redacted at Partial) | document keys, statements, query parameters, search queries |
| metadata (`<md>`, redacted at Full) | bucket, scope, collection, index, design-doc and view names |
| system data (`<sd>`, redacted at Full) | endpoints, management URIs |

Not redacted, deliberately: `ClientContextId`, which support correlates against server logs, and the server-authored `Message` / `Errors` text.

That second one is a real limit rather than a formality, and it is worth stating plainly because it bounds what this PR can claim. A N1QL error routinely quotes the statement it failed on, and `QueryErrorContext.Message` is built straight from it (`GetErrorMessage` returns `$"{error.Message} [{error.Code}]"`). So a customer at `Partial` gets `Statement` and `Parameters` tagged while the same statement text can still reach them through `Message` and `Errors`. Redacting server text would cost the diagnostic that makes the context worth having, and Java leaves it raw too — but the release note has to say so rather than let "statements are redacted" stand unqualified.

## Rebased onto 3.10

This branch was written against the two-redactor world that NCBC-4302 (#183) has since removed, so it was rebased onto `3.10` rather than merged forward. Three things changed beyond the mechanical rename, and they are worth a reviewer's attention because none of them show up as a conflict:

- **The three new redactor holders take the concrete `Redactor`, not `IRedactor`.** `QueryClient`, `SearchClient` and `AnalyticsClient` held no redactor before this PR, and as originally written they took `IRedactor` — the convention at the time. After #183 there are zero `IRedactor` fields in `src/`, and reaching the redactor through the interface costs 9.9 ns / 56 B per call against 1.5 ns / 0 B for the concrete type, paid before any log level is consulted. Leaving them as `IRedactor` would have reintroduced that path on the query, search and analytics error paths in the same release that removed it.
- **`RedactorExtensions` lost its three `*String` mirrors.** They existed only because error-context holders were split across `IRedactor` and `TypedRedactor` and both had to read the same way. There is one type now, so `UserDataString`/`MetaDataString`/`SystemDataString` are plain instance methods on `Redactor` and the extension class keeps only `OperationKey`/`OperationKeyString` — which stay extensions so that `Core.Logging` takes no dependency on `Core.IO.Operations`. The `?.ToString()` in the mirrors went with them; it was only ever needed because `IRedactor.UserData` returns `object?`.
- **One commit was dropped as obsolete.** *"Use a real redactor in tests that build error contexts"* replaced the `Mock<IRedactor>` in `CollectionManagerTests` and `ViewClientTests`, whose methods return null and silently blank every field they are meant to check. #183 did that independently and across the whole suite, so the commit had nothing left to do.

### Second rebase, onto NCBC-4213 (#186)

Rebased again on 2026-09-18 onto `b6215bc89`, because #186 landed on `3.10` and touches the same
files. The diff is four lines smaller than before: `3.10` now supplies `SearchClient`'s redactor
field itself, so this PR no longer adds it.

Worth knowing if you replay this yourself. Git reported conflicts in three **test** files and
auto-merged `src/Couchbase/Search/SearchClient.cs` **with no conflict at all** — into code that does
not compile. #186 and this branch each added `private readonly Redactor _redactor;` and the
constructor parameter at *different offsets*, so git saw two independent insertions rather than one
collision and kept both (CS0102). The conflicts git does flag were the harmless ones. The resolution
keeps `3.10`'s field ordering, so the rebased commit carries only its own 15 lines of error-context
redaction, and #186's `LogTrace` redaction survives untouched.

Verified after the rebase: 3107 unit tests on net8.0 and net10.0, 94 DI, 9 OpenTelemetry, all five
solutions build, and the warning set is byte-identical to `3.10` (`4 CS0436 / 16 CS0618 /
4 CS8632`, via `-t:Rebuild` on both sides).

## Judgement calls worth a reviewer's attention

- **Redaction happens at assignment, not at render.** This is the one real design choice, and it is the fail-closed one. The stored value is tagged, so every consumer is covered — `ToString()`, a customer's own field logging, a debugger watch window — including consumers that do not exist yet. Render-time redaction is fail-open by comparison: every render path has to be redaction-aware, and a missed one leaks silently. This PR found two such paths already. `ToString()` escaped its tags into uselessness, and the now-deleted `ToJsonString()` had the same bug unnoticed for a year — under render-time redaction that would have been a live leak rather than a cosmetic one. NCBC-4297(c) will add a third when it appends `Context` to `Message`.

  The cost is that the raw value is not recoverable from the context. Nothing in the SDK reads these fields back — the only internal readers are `Status` and `Errors`/`Code` — and for someone who enabled redaction for compliance, an unrecoverable key is closer to the intent than to a loss. If that ever needs to change, an internal raw field alongside the public tagged property is additive.
- **There is no spec for any of this, and the other SDKs do not settle it.** Redaction appears once in all 45 SDK RFCs, as an aside about a tracing tag in 0035; the `<ud>`/`<md>`/`<sd>` convention comes from the server side. Go checks the redaction level at the call site but never redacts an error context at all. Java wraps at assignment and renders in `RedactableArgument.toString()`, which is instructive rather than precedent: a 2019 "exact syntax yet to be determined" in that one method leaves `redactMeta` and `redactSystem` emitting untagged values to this day, at every call site, while every call site still looks correct. Centralised rendering makes one wrong decision invisible everywhere.
- **Java classifies identically but only ever emits `<ud>`.** `ReducedKeyValueErrorContext` uses `redactUser` for the document id and `redactMeta` for bucket, scope and collection, exactly as the table above does, but its render step returns meta and system values raw. .NET has emitted `<md>` and `<sd>` in log lines since NCBC-3079 in 2021, so the table follows .NET's own established convention, not Java's behaviour. Worth knowing before comparing the two.
- **Null and empty still pass through untouched, for a narrower reason than before.** NCBC-4302 taught the redactor not to tag a null at all, so the `*String` helpers no longer have to prevent a `<ud></ud>`; what they still have to prevent is the *rendering*, since `Redacted<T>` renders a null value as `""` and would turn an absent context field into a present, empty one. Empty is short-circuited separately, because tagging it yields a useless `<ud></ud>`. `GetClusterMap` passes `string.Empty` as the bucket name, so this is reachable.
- **`SelectBucket` is the one operation whose `Key` is not a document key** — it holds the bucket name, so it is metadata; as user data it would be stripped at Partial, losing a diagnostic Couchbase treats as safe there. The rule lives in one helper (`RedactorExtensions.OperationKey`) that **all thirteen** sites route through: the three error contexts (`ClusterNode`, `ResponseStatusExtensions`, `RetryOrchestrator`), the timeout message in `ThrowHelper`, the six `ClusterNode` log lines, and three in `RetryOrchestrator`. Previously the log line and the message classified that same value as user data while the context called it metadata.

  Routing them all through it turned up **three live leaks that pre-date this PR**, all fixed here. `ClusterNode`'s circuit-breaker line redacted the document key as *system* data (`_redactor.SystemData(op.Key)`), and system data is not tagged at `Partial` — the level whose entire purpose is stripping user data — so the key went to the log raw. Then two lines passed `op.Key` with **no redactor at all**, so the key was raw at every level including `Full`: the NMVB config-push line in `ClusterNode` (found by Copilot) and `CouchbaseBucket`'s vbucket-mapping line, which runs for *every KV operation* rather than on one NMVB path. Its endpoint was untagged too, where every sibling line uses `SystemData`.

  Worth recording how the second pair was missed the first time, because the first sweep looked thorough and was not: it grepped for a *misclassified* redaction, `redactor.X(op.Key)`, which by construction cannot match a key that reaches a log with **no** redactor on it. Two sweeps are needed, and both are clean now — no redaction of `op.Key` outside the helper, and no `op.Key` reaching a log or message without one. By subject these are log-redaction gaps and belong with NCBC-4213's rather than here; they are taken because the helper reduces each to one line, and because leaving them would leave its documented invariant false.
- **The redaction tags are restored after serializing, not produced by a relaxed encoder.** System.Text.Json escapes `<` and `>` by default, which turns every tag into an escape sequence — semantically correct and operationally useless, since `cblogredaction` matches the tags textually. The obvious fix, `UnsafeRelaxedJsonEscaping`, was the wrong way round: it unescapes *every* value in the document, and these contexts carry content the SDK neither authors nor redacts at any level — `ManagementErrorContext.Message` is the management response body, which ns_server can return as HTML, and the query and search contexts carry server-authored error text. An application embedding a rendered context in a diagnostics page would have gained an XSS sink from it, **whether or not redaction was ever enabled**. Copilot flagged this and was right.

  So `ToString()` serializes with the default HTML-safe encoder and then restores only the six tokens the redactor emits, each paired with its escaped form as taken from `JavaScriptEncoder.Default` rather than hard-coded. A document with no escaped `<` returns unchanged after one scan. Everything else — server text, markup, non-ASCII — keeps the escaping it has always had.

  This also removed the per-type `static readonly JsonTypeInfo` the four contexts had to bind from redaction-safe options: they are back on `SerializerContext.Default.X`, so there is no longer a type initializer that can throw from inside an exception's `ToString()`, and `InternalSerializationContext` and `ManagementSerializerContext` revert to their `3.10` contents exactly.
- **`EventingFunctionErrorContext` needs nothing.** It is never constructed anywhere in the SDK, and its only non-`Message` field is `[JsonIgnore(Always)]`. Because an exemption like that is permanent, each one now records the string fields its reason was checked against and `ExemptTypesStillHaveNothingToClassify` pins them, so a field added later cannot slip in through the escape hatch the table exists to close.
- **`ManagementErrorContext.Statement` is one classification over a composite value.** It holds a management URI, whose host is system data while the bucket, scope, collection and index names in its path are metadata. Calling the whole thing system data costs nothing while metadata and system data redact at the same level and neither at `Partial`; it would cost something the day a management URI carries a document key or a user name, since the URI would then stay raw at `Partial`. No test can see that coming, so it is written down at the classification table instead.

## Behaviour change

For anyone who has enabled redaction, error-context fields hold the tagged value. The affected properties are all `[InterfaceStability(Level.Uncommitted)]`, and nothing inside the SDK reads them back — the only internal consumers read `Status` (`ClusterContext`, `CouchbaseBucket`) and `Errors`/`Code` (transactions' `ConvertQueryError`), neither of which is redacted.

There is **no second behaviour change**. An earlier revision of this PR switched `ToString()` to the relaxed JSON encoder, which changed the output for everyone whether or not they had enabled redaction; that is gone. Only the six redaction tokens are unescaped now, and they cannot appear unless redaction is on, so a caller at the default `RedactionLevel.None` sees byte-identical output to `3.10`.

### Release note

Still owed by hand, and it lands in the same release as NCBC-4302's, which is also owed — the two should be written together, since both describe redaction changing in 3.10.0. Draft text for `sdk-release-notes.adoc`, under `==== New Features and Behavioral Changes`:

```asciidoc
* https://couchbasecloud.atlassian.net/browse/NCBC-4296[NCBC-4296]:
Exception error contexts now honour `ClusterOptions.RedactionLevel`, which previously applied to the log stream only.
If you have enabled redaction, error-context properties hold the tagged value: `ex.Context.DocumentKey` returns `<ud>the-key</ud>` rather than `the-key`, and the raw value is no longer recoverable from the context.
Document keys, statements, query parameters and search queries are tagged `<ud>`; bucket, scope, collection, index, design-document and view names `<md>`; endpoints and management URIs `<sd>`.
Text authored by the server is deliberately left raw, so a statement that a query error quotes back can still appear in `Context.Message` and `Context.Errors` even at `Partial`: redaction covers the fields the SDK populates, not the text the server returns.
`ClientContextId` is left raw for the same reason it always was, so that it can be correlated against server logs.
This covers the `couchbase://` path only; `couchbase2://` is tracked separately as https://couchbasecloud.atlassian.net/browse/NCBC-4300[NCBC-4300].
`Context.ToString()` renders the tags as literal angle brackets so that `cblogredaction` can match them; every other character keeps its usual JSON escaping, so output is unchanged for anyone who has not enabled redaction.
```

## Known gap: the couchbase2:// path is not covered

`GenericErrorContext` on the Stellar path has the same defect and is untouched here. `StellarRetryHandler` copies the gRPC `ResourceInfo` detail into `Fields["ResourceName"]` raw, and when the sibling `ResourceType` is `"document"` that value **is** the document key — so `DocumentNotFoundException` and `DocumentExistsException` over `couchbase2://` still write keys unredacted into the log. This is a real gap, not a scoping technicality — tracked as NCBC-4300, targeting 3.10.0.

It is left out because it is a different fix rather than more of the same:

- **No redactor exists on that path.** `StellarRetryHandler` holds no redactor, and the `IRedactor` on `StellarCluster` has never been invoked by Stellar code. It needs DI and constructor changes, not a classification change.
- **Different serialization problem.** `GenericErrorContext.ToString()` serializes an untyped `Dictionary<string, object>` reflectively, with no serializer context and no options — so it needs the relaxed-encoder fix separately, and raises a trimming/AOT question the four source-generated contexts did not.
- `ResourceName`'s classification depends on the sibling `ResourceType` value, the same value-provenance shape as `SelectBucket`'s `Key`, so the helper added here is the right model for it.
- The classification test below cannot reach an untyped bag; covering it needs per-key classification.

`ResourceType` `"path"` (a subdoc path) also has no classic equivalent in any context, so whether it is user data or metadata is an open question for that ticket, and probably a cross-SDK one.

## Testing

36 new tests. `ErrorContextRedactionTests` covers None/Partial/Full and tag survival through `ToString()`. `ErrorContextRedactionClientTests` drives the query, search, analytics, view and management paths end-to-end through mocked HTTP failures, so every context type has a site pinned. `OperationKeyRedactionTests` pins the `SelectBucket` rule in both the context and the timeout message, plus the counterparts proving an ordinary `Get` is still user data.

`ErrorContextClassificationTests` is the one worth understanding. Redaction applied field by field has exactly one failure mode — a field nobody classified — and value-based tests cannot catch it, because a new unredacted field simply is not asserted on. So every string field on all six context types is declared as either redacted or deliberately raw, and checked two ways: a reflection test that fails until a new field is listed, and a behavioural test that drives each real construction path at `Full` and asserts redacted fields come back tagged and raw ones do not. 22 of the 31 string fields are populated by a real path; the rest are covered by the reflection half. A third test scans the assembly for `IErrorContext` implementers, so a whole new context type cannot arrive unlisted either — the two deliberately left alone, `EventingFunctionErrorContext` and Stellar's `GenericErrorContext`, are named there with the reason each needs nothing.

A fourth, `EveryContextThatRendersEmitsLiteralTags`, renders every classified context that overrides `ToString()` and asserts its tags come out literal. Each context binds its own static `JsonTypeInfo` in a field initializer that reaches into a serializer context's `Default`, and both halves of that are per type — a context that misses the binding escapes its tags to `\u003C`, which still parses as JSON while defeating `cblogredaction`, and one that initializes in the wrong order throws `TypeInitializationException` from an error path. Only `KeyValueErrorContext` was rendered by a test before. The set is found by reflection, so `QueryErrorContext` and `AnalyticsErrorContext` join it the day NCBC-4297(c) gives them a `ToString()`.

**3108 pass on net8.0 and net10.0**, zero failures, plus the 94 DI and 9 OpenTelemetry tests the build-and-test gate does not run. All solutions build — `couchbase-net-client`, `-onpremise`, `-di`, `-otel` and `couchbase-fit-performer` — and the warning set is byte-identical to `3.10` leg by leg. (`couchbase-net-client-txn.sln` does not load, on this branch and on `3.10` alike: it references `src/Couchbase.Transactions`, which no longer exists.)

Both remaining items are now done. CI is green on all four legs, including the net48 Windows one that macOS cannot run. FIT [`op-onprem-func-lite` on 8.5-stable](https://github.com/couchbase/couchbase-net-client/actions/runs/35126804747) came back **4057 pass / 419 skip / 4 fail**, and all four failures are on master's nightly for the same day, so none is ours. The performer sets `RedactionLevel.Partial` unconditionally, which makes that run a real exercise of this change rather than only a regression net: `SelectBucket` was the one opcode logging an untagged key, in all 2564 of its log lines, and a rendered `QueryErrorContext` showed `Statement` and `Parameters` tagged with `ClientContextId` raw and the tags as literal angle brackets.

That run was against `c99dbfb13`, so a [**second run**](https://github.com/couchbase/couchbase-net-client/actions/runs/35251777184) covers the current head `b3975f96c` — the performer image `ghcr.io/couchbase/dotnet-fit-performer:NCBC-4296` was confirmed to carry that commit before dispatch, by reading `org.opencontainers.image.revision` off the published config blob rather than trusting the tag.

**4058 pass / 419 skip / 4 fail / 2 err — 99.85%**, and the failure set is *identical, test for test*, to master's own nightly on the same cluster that morning ([run 35165776503](https://github.com/couchbase/couchbase-net-client/actions/runs/35165776503), `8.5-stable·dotnet:main·functional`, 4042/427/4/2):

| | branch | master nightly |
|---|---|---|
| `AppTelemetryE2ETest.validateAllTypeRequestMetric` | ❌ | ❌ |
| `SingleQueryConfigTest.adhocFalse` | ❌ | ❌ |
| `ZoneAwareReadFromReplicaUnbalancedTest.unbalancedGetAllReplicasWithSelectedNotInGroup` | ❌ | ❌ |
| `ZoneAwareReadFromReplicaUnbalancedTest.unbalancedLookupInAllReplicasWithSelectedNotInGroup` | ❌ | ❌ |
| `CustomMetadataCollectionTest.failTransactionAtReplaceRemoveStaging_…` ×2 | 💥 | 💥 |

The two errors are the same transactions method at two hook points — `AFTER_STAGED_REPLACE_COMPLETE` and `AFTER_STAGED_REMOVE_COMPLETE` — both aborting with `TransactionOperationFailedException` / `DocumentNotFoundException` where the test expects the transaction to survive the metadata collection being deleted and recreated. Same two hook points in master's nightly. Nothing in this run is absent from master's, so none of it is ours. `QueryExpiryTest.expiresAtExpectedTime`, which errored in the first run, passed in this one — a flake either way, and it is absent from master's nightly too.

Worth noting for anyone reading the badge: these runs exit non-zero on any test failure, so the job shows as **failed** even when it matches master exactly. Judge it on the counts, and count 💥 as well as ❌ — transactions failures surface as 💥 and are easy to miss.

## Reading the commits

Fourteen commits. The first three are the fix. The rest respond to review: the operation-key helper, the classification tests, two cleanups — one home for the redaction-safe serializer settings, and one copy of the HTTP fixture helpers — a pass that deletes `CouchbaseException.ToJsonString()`, a test that renders every context rather than only the KV one, the closing of the classification table's escape hatch, a commit routing the operation-key redactions through one helper, the swap of the relaxed JSON encoder for a six-token restore pass, and the two raw operation keys that reached a log with no redactor at all.

That method serialized a context with the default HTML-escaping encoder, so its tags would have come out as escape sequences. It goes, along with `InterfaceRuntimeTypeConverter`, its only consumer.

On the evidence for "no caller": `AssemblyInfo.cs` names **twelve** friend assemblies, not two, and `Couchbase.FitPerformer` — the performer its comment pointed at — is one of them, so it could have seen the method even though it does not use it. Nothing in this repo calls it, including `fit/`, and the stale out-of-tree `transactions-fit-performer` checkout carries the definition with no caller. What an in-repo grep cannot see is `Sdkd`, `SdkdConsole` and `Couchbase.Transactions.FitPerformer`, which are built elsewhere; **sdkd-net has not been checked.** Flagging that rather than letting the earlier, wrong claim stand.

## Follow-up recorded on NCBC-4297

`QueryErrorContext` and `AnalyticsErrorContext` have no `ToString()` override, because both hold `List<Error>` and `Error` carries a `[JsonExtensionData] IDictionary<string, object>` that can contain Newtonsoft `JToken`s — not serializable through the source-generated context, which is why their `DebuggerDisplay` routes through `SerializeWithFallback` under `[RequiresUnreferencedCode]`. That attribute cannot go on a `ToString()` override.

This must be fixed before NCBC-4297(c) appends `Context` to `Message`, or every query and analytics exception message would end in a bare type name. The ticket's (c) currently assumes all contexts already render as compact JSON.


src/Couchbase/Core/ClusterNode.cs
```diff
@@ -438,14 +438,15 @@ public async Task<BucketConfig> GetClusterMap(ConfigVersion? latestVersionOnClie
                 configOp.Revision = latestVersionOnClient.Value.Revision;
             }
 
+            var redactor = _redactor;
             var config = await ExecuteInternalOperationAsync(ConnectionPool, configOp,
                 ExecuteOpImmediatelyAsync,
-                static (status, op) =>
+                (status, op) =>
```
jeffrymorris: The missing static to capture redactor will make the previously cached delegates into new closure allocations per call. It's a performance regression.

src/Couchbase/Core/IO/ResponseStatusExtensions.cs
```diff
@@ -76,23 +77,25 @@ public static bool IsRetriable(this ResponseStatus status, IOperation op)
             }
         }
 
-        public static Exception CreateException(this ResponseStatus status, IOperation op, IBucket bucket) => CreateException(status, op, bucket.Name);
+        public static Exception CreateException(this ResponseStatus status, IOperation op, IBucket bucket,
+            TypedRedactor redactor) => CreateException(status, op, bucket.Name, redactor);
 
-        public static Exception CreateException(this ResponseStatus status, IOperation op, string bucketName)
+        public static Exception CreateException(this ResponseStatus status, IOperation op, string bucketName,
+            TypedRedactor redactor)
```
jeffrymorris, line 84: I feel like passing in a specific IRedactor implementation will make testing/maintenance more difficult in the future.

src/Couchbase/Core/Logging/TypedRedactor.cs
```diff
@@ -45,6 +45,22 @@ public Redacted<T> SystemData<T>(T message)
...
+
+        public string? UserDataString(string? value) =>
+            string.IsNullOrEmpty(value) ? value : UserData(value).ToString();
+
+        public string? MetaDataString(string? value) =>
+            string.IsNullOrEmpty(value) ? value : MetaData(value).ToString();
+
+        public string? SystemDataString(string? value) =>
+            string.IsNullOrEmpty(value) ? value : SystemData(value).ToString();
+
```
jeffrymorris, line 63: These methods feel cringy/bloated. We already have the IRedactor interface and then these go around it.

src/Couchbase/Core/ClusterNode.cs
davidkelly: Good catch — that was a real regression, and it's fixed in the last commit ("Keep the internal-operation projectors static").

Rather than capturing the redactor, `ExecuteInternalOperationAsync` now hands it to the projector as a third argument:

```csharp
Func<ResponseStatus, TOperation, Redactor, TResult> projector
...
return projector(status, operation, _redactor);
```

So all three sites are back to `static (status, op, redactor) => …` and the compiler caches the delegate again — no closure, no per-call allocation. The one site that doesn't need it just discards the parameter (`static (status, op, _) =>`).

src/Couchbase/Core/IO/ResponseStatusExtensions.cs
davidkelly: Worth flagging first that this line now reads `Redactor` rather than `TypedRedactor` — the branch has been rebased onto NCBC-4302 (#183), which collapsed the two redactors into one and moved every internal call site onto the concrete type. So it's no longer a one-off choice in this PR; it's the convention that PR established.

To answer the substitution question directly, though: **you can't put a mock in, and that's now enforced rather than discouraged.** `Redactor` is `internal sealed` with non-virtual methods, so Moq has nothing to proxy, and `sealed` means no type can satisfy `where TImplementation : TService` on either `AddClusterService` overload — that was verified during #183 rather than assumed. I'd rather state that plainly than have it discovered later.

What tests do instead is construct a real one: `new Redactor(RedactionLevel.None)` for pass-through, `Partial`/`Full` to assert on tags (there's a `TestRedactor` helper holding all three). In practice that's been better than the mocks it replaced — `Mock<IRedactor>`'s methods return null, so any field redacted through one silently became null instead of the value the test meant to check. That bit two tests in this PR, which is why #183 deleted every one of them. It also pushes these tests toward asserting on output rather than on interactions, which is the assertion you actually want here.

What's genuinely lost is interaction verification and canned returns. If that turns out to matter, the cheap way back isn't the interface — it's a nullable hook field on the concrete class. Measured over the real `Redacted<T>`, 200M iterations: sealed concrete **1.52 ns / 0 B**, concrete + hook **1.83 ns / 0 B**, generic interface **3.66 ns**, today's `object`-based `IRedactor` **9.93 ns / 56 B**. So ~0.3 ns buys substitution back with no allocation, where the interface costs 2.4–6.5x. Happy to raise that as its own ticket if you want it — it just shouldn't ride along with this fix.

src/Couchbase/Core/Logging/TypedRedactor.cs
davidkelly: This comment is what prompted NCBC-4302, so thanks for pushing on it — and the rebase onto #183 has removed the part you were reacting to.

There's no longer a second set of methods going around the interface. The `IRedactor` extension mirrors are deleted; what's left is three instance methods on the single `Redactor` class. (`IRedactor` itself is untouched and still public — it's just no longer used internally.)

On whether the three should exist at all: they're carrying **67 call sites across 12 files**. The error contexts store plain `string`, so without them each site writes `redactor.UserData(x).ToString()` plus its own null/empty guard — and the guard is the whole point. `Redacted<T>` renders a null as `""`, which would turn an absent context field into a present, empty one, and tagging an empty value gives a useless `<ud></ud>`. `GetClusterMap` passes `string.Empty` as the bucket name, so it's reachable rather than theoretical. Sixty-seven chances to forget it felt worse than three small methods.

Open to a better name if `…String` is the part that grates.

src/Couchbase/Core/RedactionSafeJson.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Callers hold the result in a static field, created on demand rather than in a field
+        /// initializer: a serializer context's <c>Default</c> is not yet constructed while that
+        /// context is running its own static initialization.
+        /// </remarks>
+        public static JsonSerializerOptions Create(JsonSerializerOptions source) =>
+            new(source)
+            {
+                Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
```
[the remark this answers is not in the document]
davidkelly: Agreed, and fixed in 4be6e3bd — the trade was the wrong way round.

The tags are six known tokens, so there was never a need to unescape the other several hundred characters to reach them. `ToString()` now serializes with the default HTML-safe encoder as before and restores only `<ud>`, `</ud>`, `<md>`, `</md>`, `<sd>` and `</sd>` afterwards. Each token's escaped form comes from `JavaScriptEncoder.Default.Encode(...)` rather than being hard-coded, so the pairing can't drift if System.Text.Json ever changes how it escapes an angle bracket. A document with no escaped `<` returns unchanged after a single scan, which is every context when redaction is off.

Your framing of the risk is the part worth keeping on the record: it applied **whether or not redaction was enabled**, because none of the content in question is redacted at any level. `ManagementErrorContext.Message` is the raw ns_server body and the query and search contexts carry server-authored error text, so the relaxed encoder was handing back unescaped markup to callers who had never opted into anything.

Two things fell out of it that weren't the point but are worth having:

- **The per-type `static readonly JsonTypeInfo` is gone.** It only existed to bind each context to the relaxed options. All four contexts are back on `SerializerContext.Default.X`, so there is no longer a type initializer that can throw from inside an exception's `ToString()` — a separate review comment on this PR.
- **`InternalSerializationContext` and `ManagementSerializerContext` revert to their `3.10` contents exactly**, and non-ASCII is escaped again. The PR's "one thing changes for everyone, redaction or not" section is now empty: at the default `RedactionLevel.None` the output is byte-identical to `3.10`.

Pinned by `RestoringTagsLeavesServerContentEscaped`, which drives the management path with `</script><script>alert(1)</script>` as the response body and asserts both halves together — the `<sd>` tag comes back literal and the markup does not. Replacing the restore with a faithful stand-in for the relaxed encoder fails that case **alone**: the other 42 error-context tests, including all four `EveryContextThatRendersEmitsLiteralTags` cases, still pass. That is precisely the blind spot you were pointing at.

src/Couchbase/Core/Logging/RedactorExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <summary>
+        /// Redacts an operation's key with the classification that key actually carries.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="OpCode.SelectBucket"/> is the only operation whose <see cref="IOperation.Key"/>
+        /// is not a document key - it holds the bucket name. Bucket names are metadata, so tagging
+        /// it as user data would have it stripped at <see cref="RedactionLevel.Partial"/>, losing a
+        /// diagnostic that Couchbase treats as safe at that level. Every site that redacts an
+        /// operation key routes through here so that the log line, the exception message and the
+        /// error context cannot classify the same value differently.
```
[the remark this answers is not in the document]
davidkelly: You're right, and the invariant claim was mine to get wrong. Fixed in 81767598.

The commit before yours had routed the *misclassified* redactions through the helper — including a real one, where `ClusterNode`'s circuit-breaker line tagged the document key as **system** data, which isn't redacted at `Partial` — and then claimed the invariant held everywhere. It was checked by grepping for `redactor.X(op.Key)`, which by construction **cannot match a key that reaches a log with no redactor on it at all**. The sweep looked thorough and was blind to exactly the case you found.

Sweeping for the raw form instead turned up a second one, and it's the worse of the two:

| | |
|---|---|
| `ClusterNode.cs:724` | the NMVB config-push line you flagged — one path, debug only |
| `CouchbaseBucket.cs:325` | `"Mapping op {OpCode} to {Endpoint} for key {Key} …"` — the **vbucket-mapping line, so every KV operation** |

Its `endPoint` argument was untagged too, where every sibling line in `ClusterNode` uses `_redactor.SystemData(...)`. Both now go through `RedactorExtensions.OperationKey`, and the endpoint is tagged as system data.

Both sweeps are clean now: no redaction of `op.Key` outside the helper, and no `op.Key` reaching a log or message without one.

For the record on scope — both of these pre-date this PR and are log-redaction gaps rather than error-context ones, so by subject they belong with NCBC-4213's follow-up. They're taken here because the helper this PR adds reduces each to one line, and because leaving them would leave the helper's documented invariant false, which is the thing you were objecting to.

src/Couchbase/Core/RedactionSafeJson.cs
```diff
@@ -0,0 +1,92 @@
...
+        public static string RestoreTags(string json)
+        {
+            if (json.IndexOf(EscapedLessThan, StringComparison.Ordinal) < 0)
+            {
+                return json;
+            }
+
+            foreach (var (escaped, literal) in Tags)
+            {
+                json = json.Replace(escaped, literal);
```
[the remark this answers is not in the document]
davidkelly: Confirmed and fixed in c0899cef5. The corruption reproduces exactly as you describe.

A field holding the literal text `\u003Cud>` serializes to `{"message":"\\u003Cud\u003E"}`, the match begins at the second backslash, and the replacement consumes the `\\` pair to leave `{"message":"\<ud>"}` — `\<` is not a valid JSON escape, so `Context.ToString()` returns a document that will not parse. That is the string people paste into a ticket, so it matters more than its likelihood suggests.

`RestoreTags` now walks the document rather than calling `String.Replace`. Backslashes pair off inside a JSON string, so only an odd-length run ends in an escape introducer, and a tag is restored only when its leading backslash is one. No allocation when nothing matches, and the fast-path guard is unchanged.

Two things I want on the record rather than quietly folded in.

**The second half of your comment is right and is not fixed**, because I do not think it can be. A literal `<ud>` arriving in server text is unescaped here too, at every level including `None`. Telling it apart from a tag the redactor added needs a marker the redactor emits only for this path — but the same `Redacted<string>` is rendered straight into log messages, where the tag must already be literal, so there is no separate representation to key on. Building one is a larger change than this defect justifies, which is why I have taken the parity fix and left this documented in the remarks instead.

What it does falsify is my own claim on the previous round that output at `None` is "byte-identical to `3.10`". It is byte-identical *unless the payload contains a literal tag*. The remarks now say that rather than saying nothing changes when redaction is off.

**A note on the test, since it is a trap.** The body has to be written `"\\u003Cud>"`. C# applies `\uXXXX` escapes in the lexical phase, before string literals are formed, so `@"\u003Cud>"` is four characters — `<ud>` — and so is the raw-string form. A test written the obvious way would have passed against the broken implementation while asserting nothing. `RestoringTagsLeavesTheDocumentParseable` fails on the previous implementation; I checked by reverting rather than assuming.

src/Couchbase/Core/Logging/RedactorExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <para>
+        /// Unlike the <c>*String</c> helpers this does not special-case an empty key: internal
+        /// operations leave <see cref="IOperation.Key"/> at <see cref="string.Empty"/> and have
+        /// always logged as an empty tag.
+        /// </para>
+        /// </remarks>
+        public static Redacted<string> OperationKey(this Redactor redactor, IOperation op) =>
+            op.OpCode == OpCode.SelectBucket
+                ? redactor.MetaData(op.Key)
+                : redactor.UserData(op.Key);
```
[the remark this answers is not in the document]
davidkelly: Confirmed, and your structural point is the one I have acted on. Fixed in 9a767cf8e.

`GetCid.Key` is as you describe: `CouchbaseCollection.cs:1367` assigns the fully qualified name on the `sendAsBody: false` branch, and `GetCidWithFallbackAsync` takes that branch on `InvalidArgumentException`, so it is live rather than dead code.

Rather than add the two opcodes you named, I swept every assignment to `IOperation.Key` in the SDK, because adding what a reviewer noticed is how the helper got into this state. That found one more, and it is the significant one:

| | Key holds | Reaches the helper? |
|---|---|---|
| `SelectBucket` | bucket name | yes |
| **`Helo`** | connection id + client description | **yes — every connection, via `ClusterNode.ExecuteOp`** |
| `GetCidByName` | `scope.collection`, pre-7.0 fallback | yes |
| `GetSidByName` | scope name | no — nothing constructs a `GetSid` |
| `SaslList`/`SaslStart`/`SaslStep` | mechanism name | no — own send path, bypasses `ExecuteOp` |

HELO is the one that was actually costing something: it runs on every connection, so at `Partial` the SDK identifier and connection id were being stripped out of every negotiation log line. That is a wider loss than either case in your comment.

All seven are now named in a switch, with the remarks recording that the list came from the `Key =` sweep rather than from review, so the next reader can tell whether it is exhaustive. `GetSidByName` and the SASL opcodes are classified even though they do not reach the helper today, so the first caller to change that does not have to remember.

**On scope**, and I would rather say this plainly than keep absorbing it. This is log-redaction work, not error-context work, and by subject it belongs with NCBC-4213's follow-up — the same was true of the two operation-key fixes in the previous round. I have taken it here because the helper is this PR's and the invariant it documents was false without it. But that is now three review rounds pulling log-redaction scope into a change whose stated subject is error contexts, and the PR is 137 files. Anything further of this kind gets filed against the NCBC-4213 follow-up rather than added here.

tests/Couchbase.UnitTests/Core/Exceptions/ErrorContextClassificationTests.cs
```diff
@@ -0,0 +1,535 @@
...
+        /// and a plain <see cref="string.Replace(string,string)"/> cannot tell a backslash that
+        /// introduces an escape from one that is itself an escaped literal backslash: a field
+        /// holding the text <c>\u003Cud&gt;</c> serializes as <c>\\u003Cud\u003E</c>, where the
+        /// match begins at the second backslash and leaves <c>\&lt;ud&gt;</c> - not a valid JSON
+        /// escape, so the whole context stops parsing. Reported by review on NCBC-4296.
+        /// <para>
+        /// The body has to be written with an escaped backslash. C# applies <c>\uXXXX</c> escapes
+        /// in the lexical phase, before string literals are formed, so a verbatim or raw literal
+        /// does not hold this text either - <c>@"\u003Cud&gt;"</c> is four characters, and a test
+        /// written that way asserts nothing.
```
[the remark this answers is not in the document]
davidkelly: ignoring this one for now

jeffrymorris (approved):
LGTM

PR #184, opened 2026-09-16
NCBC-4233: Stop running the cleanup timer callback as async void

davidkelly:
## Summary

`PerCollectionCleaner`'s `Timer` callback was `async void` — the last one in `src/`. Anything escaping it was re-thrown on a ThreadPool thread with no `SynchronizationContext` to catch it, which terminates the process.

The ticket framed this as future-proofing, but it was a **live crash path**. The body's catch-all does not cover it: an exception thrown inside a `catch` block is not caught by a sibling `catch`, and two of the handlers go on to do real work — the collection-not-found handler invokes a caller-supplied callback, and both it and the auth-error handler `await DisposeAsync()`.

Tracing that surfaced three more defects in the same class, all fixed here:

- Every handler logged **before** releasing the timer-callback mutex, so a throwing log sink leaked it. `DisposeAsync` waits on that mutex without a timeout, so it then blocked forever, hanging `Transactions.DisposeAsync`.
- A throwing sink likewise escaped a handler that had just decided to stop, skipping the teardown — trading a crash for a cleaner that retries a disowned collection once per window, forever.
- `DisposeAsync` guarded on an unsynchronised `IsCancellationRequested` check. Two callers race it (the `LostTransactionManager` at shutdown, and a cycle disposing itself on an auth error or a deleted collection), and the second could return while the first was still mid-disposal — so the cluster was disposed out from under `RemoveClient` and the client record left behind, keeping this client's share of ATRs assigned to a client that is never coming back.

## Changes

- **Split the callback** into a `void` adapter and an `async Task` method whose outer catch cannot fault — which is what makes discarding the returned Task honest rather than fire-and-forget.
- **Mutex released from a `finally`, teardown run from an outer one.** Handlers now only record *why* the cycle is stopping, so a throw on their way out — a logging provider disposed during shutdown is the likely one — cannot cost us either invariant.
- **Disposal runs once through a `Lazy<Task>`** that every caller awaits. It yields first, so nothing runs under `Lazy`'s lock while `Stop()` cancels a token whose registrations execute inline.
- **The whole teardown path survives a dead log sink.** `DisposeOnceAsync` logs between each of its steps, so those go through a `TryLog` that swallows — otherwise the first throw skips the rest, including the client-record removal. `DisposeOnceAsync` is also total, because `_disposal` caches its Task and awaiting a faulted Task rethrows at every later caller. A teardown that does not complete is reported at Warning, not lost among the breadcrumbs.
- **Timer built from the injected `TimeProvider`**, via a new `TimerFactory.CreateWithFlowSuppressed` overload. Makes the callback drivable from a `FakeTimeProvider`, and stops this timer pinning the `ExecutionContext` present at lazy bootstrap — the leak every other long-lived periodic timer in the SDK already avoids, and which this one was the sole holdout on.
- **The constructor never starts the timer.** `TestHooks` is assigned by an object initializer *after* it returns, and `TestHooks.BeforeGetRecord` is the first call the first cleanup cycle makes — so a self-starting cleaner ran its first pass against the default hooks. `CleanerForCollection` now constructs, wires up, then calls `Start()`. `startDisabled` is gone from `PerCollectionCleaner`, `LostTransactionManager` and the FIT performer.
- **Drive-by:** deleted a `public void Dispose()` that disposed nothing (it called `Stop()`, leaving the timer, CTS and semaphore alive), and folded `TimerFactory`'s two overloads onto one shared helper.

## Testing

Thirteen tests drive the callback and the disposal path through a `FakeTimeProvider`. Each fix has one that fails without it — verified by reverting each in turn; the mutex one fails by hanging, and the constructor one by observing a cycle that ran before its hooks were wired. No wall-clock waits: a fake timer given a zero due time fires synchronously inside `Start()`, and the disposal failures are routed through a log call inside a `catch` so they escape `RemoveClient`'s retry loop rather than sitting through its delays.

3013 unit tests pass on net8.0 and net10.0. net48, netstandard2.0 and netstandard2.1 compile clean.

FIT: [`op-onprem-func-lite`](https://github.com/couchbase/couchbase-net-client/actions/runs/35461999761) (3 nodes, needed — `TransactionTestBase` fails wholesale on 1-node clusters), against performer `NCBC-4233` built from ff5a70952.

## Review notes

Review raised four things, all addressed:

- The `Lazy<Task>` caches a failed disposal and re-raises it at every later caller. Correct — though by a different mechanism than named: an `async` factory never throws synchronously, so `Lazy` caches a *faulted Task* rather than an exception. Fixed by making the factory total, and by making the logging between the teardown steps unable to throw in the first place.
- The `Task.Yield()`'s original justification did not hold. `Task.Delay(token)` does not resume its continuation inline (measured), KV operations complete through `AsyncStateBase` with `RunContinuationsAsynchronously`, and the two plain `TaskCompletionSource`s in `src/` are unreachable from cleanup. Kept for a reason that does hold: cancellation registrations always run inline on the cancelling thread, so without it `Stop()` executes them under `Lazy`'s lock.
- `TryLog` was hardcoded to Debug, which silently demoted a failed teardown that previously surfaced at Warning. Now takes a level.
- The teardown was an `await` inside a `finally` inside a `finally`. Extracted to `RunStopAsync`.

The last commit trims the comments this change added — 207 lines out, 89 back. The history and the review argument live in the commit messages rather than the source.


src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
```diff
@@ -35,30 +36,50 @@ internal class PerCollectionCleaner : IAsyncDisposable
         // ProcessClient, which the timer-callback mutex serializes - so no concurrent collection is needed.
         private readonly AtrCleanupQueue _atrsToClean;
         private readonly CancellationTokenSource _cancelToken = new ();
+        // Lazy's default thread-safety mode runs the factory once under a lock and hands every other caller
+        // the same Task - which is exactly the "dispose once, everyone awaits it" contract wanted here.
+        private readonly Lazy<Task> _disposal;
```
jeffrymorris, line 41: I am pretty sure if this throws an exception it's gonna cause an issue because it's cached and subsequent calls will rethrow the exception. Maybe add a test to prove/disapprove it?

davidkelly: Good catch — right conclusion, though by a different mechanism than the one you named, and the mechanism turned out to matter less than what it was hiding.

Lazy's own exception caching can't fire here: the factory is `DisposeOnceAsync`, an `async Task` method, and async methods never throw synchronously — so Lazy always gets a Task back and caches *that*, never an exception. But the Task it caches can be **faulted**, and awaiting a faulted Task rethrows every time. Identical observable behaviour. I probed it rather than trust my memory:

```
async factory (what we have):  first threw, second threw | IsValueCreated=True
sync-throwing factory:         first threw, second threw | IsValueCreated=False
```

The second line is real Lazy exception caching; ours is the first.

What made it worth fixing rather than just documenting is what's most likely to throw in there. `DisposeOnceAsync` had five `LogDebug` calls between its teardown steps, plus one in `Stop()`. A logging provider disposed before the SDK is the realistic way a sink throws, and disposal is exactly when that race is on — so a dead sink could skip the timer disposal and `RemoveClient()`, leaving this client's share of the ATRs assigned to a client that is never coming back. With the `Lazy`, no later caller would retry it. I'd actually conceded that gap in a comment on the collection-not-found teardown and decided it wasn't worth closing; your question makes that a much harder sell, because the `Lazy` makes the half-finished state permanent.

Fixed in 4742d963e, two parts:

1. Every log on the teardown path now goes through a `TryLog` that swallows — same precedent as `TryReleaseMutex` just above it. With nothing throwable between the steps, the sequence always runs to the end. Worth saying that an outer catch **alone** would not have fixed this: it makes the Task complete rather than fault, but the steps after the throw are still skipped, so it only trades a loud half-disposed cleaner for a quiet one.
2. `DisposeOnceAsync` is now total — the discipline `RunCleanupCycleAsync` already documents — so the cached Task can never be faulted and this is closed by construction rather than by argument. Swallowing costs nothing the callers were using: `RemoveClientEntries` already catches and logs around its `Task.WhenAll` (`LostTransactionManager.cs:130`), and the self-disposing cycle discards it into its own guard.

Three tests, each verified to fail without the part it covers:

- `ThrowingLogSink_DuringDisposal_StillRemovesTheClientRecord`
- `FailedDisposal_IsNotRethrownToLaterCallers` — disposes three times through a failed teardown; before this, all three threw
- `FailedDisposal_DoesNotFaultTheSelfDisposingCleanupCycle` — the production interleaving, a cycle disposing itself while the manager disposes it at shutdown

No wall-clock waits: the fault is routed through a log call inside a `catch`, so it escapes `RemoveClient`'s retry loop rather than sitting through its delays.

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
```diff
@@ -86,40 +109,87 @@ public override string ToString()
...
-            }
-        }
+            // Yield before touching anything. An async method runs synchronously until its first await that
+            // actually yields, and this one is a Lazy value factory - so without this, everything down to the
+            // first incomplete await executes while Lazy's recursion guard is armed. Stop() below cancels a
+            // token whose registrations run inline on this thread, which can resume a parked cleanup cycle
+            // here; if that cycle then unwinds to its own teardown (the auth path does, with no await on the
+            // way), it re-enters _disposal.Value on this very thread and Lazy throws. Yielding first hands the
+            // Task back immediately, so a re-entrant caller simply awaits the disposal already in flight.
+            await Task.Yield();
```
emilienbev, line 156: I'm not sure I understand the comment on this about the "recursion guard", but I deleted this and the tests still passed.

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
```diff
@@ -86,40 +109,87 @@ public override string ToString()
...
             }
-            else
+            catch (Exception ex)
             {
-                _logger.LogDebug("PerCollectionCleaner for '{bkt}' is already disposed.", FullBucketName);
+                // The steps themselves can still throw where TryLog cannot help: Timer.Change and
+                // CancellationTokenSource.Cancel/Dispose can raise ObjectDisposedException, Cancel surfaces
+                // anything an inline registration throws, and RemoveClient logs from inside its own catch
+                // blocks.
+                TryLog("Disposal of '{bkt}' did not complete: {ex}", FullBucketName, ex);
```
emilienbev: With the method below these incomplete disposals only log at Debug but it whereas it was WARN before. Since out debug logs are already heavy I'd probably want to add a Level parameter to TryLog below and do this one at WARN?

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
```diff
@@ -155,43 +274,88 @@ private async void TimerCallback(object? state)
...
+                {
+                    TryReleaseMutex();
+                }
             }
-            catch (Exception ex)
+            finally
             {
-                _logger.LogWarning("Processing of bucket '{bkt}' failed unexpectedly: {ex}", FullBucketName, ex);
-                TryReleaseMutex();
+                if (stop != StopReason.None)
```
emilienbev, line 340: Mother of all try/catches 😅 I would probably pull the code in that last `finally` in its own `private Task RunStopAsync(StopReason reason)` to make it easier to read and so we don't have an `await DisposeAsync().CAF();` inside a `finally` _inside a `finally`_

emilienbev (commented):
Left a couple nits/questions, but otherwise LGTM 👍 The code comments are pretty verbose, I would try to cut them down to much shorter high-level statements where necessary.

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
davidkelly: oh - that's a good point!

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
davidkelly: You're right, and the comment was wrong — thanks for pulling on it. Nothing tested it because nothing *could* trip it.

I measured the three ways a parked await can resume when `Stop()` cancels the token:

```
Task.Delay(ct)                                inline=False  no exception
TCS via ct.Register (default)                 inline=True   *** InvalidOperationException
TCS via ct.Register (RunContinuationsAsync)   inline=False  no exception
```

Line 2 is exactly the Lazy recursion-guard trip the comment describes, so the *shape* is real — but ``Task.Delay`` is not it. .NET's `DelayPromise` completes its continuations asynchronously. I checked the other candidates on this path too: KV operations complete through `AsyncStateBase`, which passes `RunContinuationsAsynchronously` (`AsyncStateBase.cs:65`), and the only two plain `TaskCompletionSource`s in `src/` (`TaskHelpers.WhenAnySuccessful`, `OperationWrapper`) aren't reachable from cleanup.

I've kept the yield, but for a reason that does still hold and that I'd missed: cancellation registrations always run inline on the cancelling thread — `RunContinuationsAsynchronously` only governs Task continuations, not registrations — so without the yield, `Stop()` executes arbitrary registration callbacks while holding Lazy's lock. The comment now says that, and records what was measured about the claim it replaces so nobody has to re-derive it.

Fixed in 920c2b42f.

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
davidkelly: Good catch, and it's worse than a level nit — it's a regression I introduced in the previous commit.

Before `DisposeOnceAsync` was made total, a failed teardown propagated out to `LostTransactionManager.RemoveClientEntries`, which logs it at Warning (`LostTransactionManager.cs:130`). Swallowing it here silently demoted a half-disposed cleaner to a Debug breadcrumb — in, as you say, an already heavy stream.

`TryLog` now takes a `LogLevel`, defaults to Debug for the breadcrumbs, and the "did not complete" message goes out at Warning. There's a test pinning it, verified to fail when the level is put back.

Worth noting the test needed to match on the message, not just the level: `RemoveClient` emits its own Warning on that same path, so asserting `Contains(Warning)` passed even with the fix reverted.

Fixed in 920c2b42f.

src/Couchbase/Client/Transactions/Cleanup/LostTransactions/PerCollectionCleaner.cs
davidkelly: Agreed — `await DisposeAsync()` inside a `finally` inside a `finally` is too much to ask anyone to hold in their head. Pulled out into `private async Task RunStopAsync(StopReason reason)` as you suggested, leaving a single `await RunStopAsync(stop)` at the call site.

Behaviour-preserving: the ordering that actually matters — the inner `finally` releasing the mutex before `DisposeAsync` re-acquires it — is unchanged, since the call still happens from the outer `finally`. I left the comment explaining *why* the teardown is deferred at the call site rather than moving it into the new method, since that's the non-obvious part.

Fixed in 920c2b42f.

emilienbev (approved):

PR #185, opened 2026-09-16
NCBC-4294: Add GetReplica for a specific replica

emilienbev:
## Motivation

The CRUD RFC (revision 23) adds `GetReplica` which allows to send a request to one replica chosen by a strategy.

## Changes

- Added `GetReplicaAsync(id, GetReplicaStrategy, GetReplicaOptions)` to `ICouchbaseCollection` marked Uncommitted. The only strategy for now is `GetReplicaStrategy.FromIndex(ReplicaIndex, options)` with a wrap option.
- The `ReplicaRead` operation carries the strategy. `CouchbaseBucket` resolves the replica from the current vBucket map on every dispatch, so each retry uses the latest config. The RFC rules decide between `ReplicaIndexOutOfBoundsException` and `ReplicaIndexCurrentlyUnavailableException`. Both fail fast and carry a `KeyValueErrorContext`.
- A `KeyNotFound` on a strategy based read maps to the new `DocumentNotFoundOnReplicaException`, which extends `DocumentNotFoundException`.
- Memcached buckets reject the strategy and Couchbase2 throws `FeatureNotAvailableException`.


fit/Couchbase.FitPerformer/gRPC/sdk.caps.proto
```diff
@@ -153,6 +153,13 @@ enum Caps {
   // The SDK supports the RESUMING status in the result of functions_status() of the eventing management API.
   SDK_EVENTING_RESUMING_FUNCTION_STATUS = 40;
 
+  // The SDK supports hybrid search score fusion (SDK-RFC 52).
+  SDK_SEARCH_SCORE_FUSION = 41;
```
jeffrymorris: Why is Hybrid search score fusion in this replica read patch? [NCBC-4272](https://jira.issues.couchbase.com/browse/NCBC-4272)

fit/Couchbase.FitPerformer/gRPC/sdk.cluster.query.index_manager.proto
```diff
@@ -7,10 +7,7 @@ option go_package = "github.com/couchbaselabs/transactions-fit-performer/protoco
 option ruby_package = "FIT::Protocol::SDK::Cluster::Query::IndexManager";
 option java_multiple_files = true;
 
-import "shared.collection.proto";
 import "sdk.query.index_manager.proto";
-import "sdk.query.index_manager.options.proto";
```
jeffrymorris: What does this have to do with replica reads?

fit/Couchbase.FitPerformer/gRPC/sdk.collection.query.index_manager.proto
```diff
@@ -7,7 +7,6 @@ option go_package = "github.com/couchbaselabs/transactions-fit-performer/protoco
 option ruby_package = "FIT::Protocol::SDK::Collection::Query::IndexManager";
 option java_multiple_files = true;
 
-import "shared.collection.proto";
```
jeffrymorris: ?

fit/Couchbase.FitPerformer/gRPC/sdk.query.index_manager.options.proto
```diff
@@ -7,8 +7,6 @@ option go_package = "github.com/couchbaselabs/transactions-fit-performer/protoco
 option ruby_package = "FIT::Protocol::SDK::Query::IndexManager";
 option java_multiple_files = true;
 
-import "shared.collection.proto";
```
jeffrymorris: ?

fit/Couchbase.FitPerformer/gRPC/shared.bounds.proto
```diff
@@ -17,9 +17,10 @@ message CounterGlobal {
 }
 
 message Counter {
-  // Each counter gets a unique id (unique to the WorkloadRunRequest, not globally unique).  The performer will create these on-demand.
-  // Each counter is bound to a WorkloadRunRequest.
-  // E.g. concurrent WorkloadRunRequests both referencing "counter1" should have separate isolated counters.
+  // Each counter is identified by its id, which the performer creates on-demand.
+  // Counters with the same id are shared across workloads and WorkloadRunRequests.  Use unique ids
+  // when you want isolated counters (e.g. per-workload count bounds), and shared ids when you
+  // want a single counter to coordinate multiple workloads (e.g. counter_eq bounds with setCounter).
```
jeffrymorris: Read from replica?

avsej:
Reviewed at [9f269475](https://github.com/couchbase/couchbase-net-client/commit/9f2694758329eca6f385aa4996fc6a52af4e21d4) against [RFC-0053 revision 23](https://github.com/couchbaselabs/sdk-rfcs/blob/36f7d3f3b8ca3a514fe04173afb7e3342055d106/rfc/0053-sdk3-crud.md?plain=1#L696-L748), which is couchbaselabs/sdk-rfcs#165 and not yet merged. Also compared against couchbase/couchbase-cxx-client#1078.

## 1. `numReplicas` is not covered as re-read per attempt

[RFC-0053 revision 23](https://github.com/couchbaselabs/sdk-rfcs/blob/36f7d3f3b8ca3a514fe04173afb7e3342055d106/rfc/0053-sdk3-crud.md?plain=1#L701) requires both inputs to be re-read on each attempt:

> Each retry should apply the implementation logic afresh, with the latest available bucket config and vbucket map.

[`Replica_Is_Reselected_On_Every_Dispatch`](https://github.com/couchbase/couchbase-net-client/blob/9f2694758329eca6f385aa4996fc6a52af4e21d4/tests/Couchbase.UnitTests/CouchbaseBucketGetReplicaTests.cs#L86-L99) moves the chain from `[1,0,2,3]` to `[1,0,3,2]` and holds `numReplicas` at 3. That covers the vBucket map. The replica count is identical on both dispatches, so a resolver that read it once and cached it would pass.

A case that changes the count, alongside it in [`CouchbaseBucketGetReplicaTests`](https://github.com/couchbase/couchbase-net-client/blob/9f2694758329eca6f385aa4996fc6a52af4e21d4/tests/Couchbase.UnitTests/CouchbaseBucketGetReplicaTests.cs):

```csharp
[Fact]
public async Task A_Grown_Chain_Satisfies_A_Previously_Out_Of_Bounds_Index()
{
    var bucket = await CreateBucketAsync([1, 0], numReplicas: 1);
    using var op = new ReplicaRead<byte[]>(Key, GetReplicaStrategy.FromIndex(ReplicaIndex.Second));

    await Assert.ThrowsAsync<ReplicaIndexOutOfBoundsException>(() => bucket.SendAsync(op, default));
    Assert.Empty(bucket.ServersThatReceivedOps);

    await bucket.ApplyVBucketMapAsync([1, 0, 2, 3], numReplicas: 3);

    await bucket.SendAsync(op, default);
    Assert.Equal([2], bucket.ServersThatReceivedOps);
}
```

It also covers the out-of-bounds boundary, and that an index error does not latch across attempts.

`couchbase-cxx-client` has the equivalent in [`revalidation_lets_a_grown_chain_satisfy_a_previously_invalid_index`](https://github.com/couchbase/couchbase-cxx-client/blob/4ffba956b4e0e57b54658559916039e4c841484f/test/unit/core/get_replica.cxx#L508-L514).

## 2. A chain entry naming a node absent from the topology diverges from C++

A vBucket map row can name a node index that the node list does not contain. [RFC-0053 revision 23](https://github.com/couchbaselabs/sdk-rfcs/blob/36f7d3f3b8ca3a514fe04173afb7e3342055d106/rfc/0053-sdk3-crud.md?plain=1#L710-L720) gives decision rules for a `-1` entry and says nothing about this one, so the two SDKs are free to differ, and they do.

`GetReplicaFromIndexStrategy.SelectReplica` tests each entry against `-1` only. Such an index is therefore returned as the replica to read from, and the caller waits out the whole timeout budget.

| | Resolver | Caller gets | Time to failure |
|---|---|---|---|
| C++ | rejects the index | `replica_index_currently_unavailable` | immediate, no network |
| Here | returns the index | `UnambiguousTimeoutException` | the full timeout budget |

C++ bounds the node index against the node list, not just against `-1`, in the [`readable` lambda](https://github.com/couchbase/couchbase-cxx-client/blob/4ffba956b4e0e57b54658559916039e4c841484f/core/impl/replica_utils.cxx#L84-L95) of `core/impl/replica_utils.cxx`:

```cpp
const auto readable = [&](std::size_t replica) -> std::optional<std::size_t> {
  if (replica >= chain_length) {
    return {};
  }
  // A -1 entry and a node index the topology does not list mean the same
  // thing here: nothing to send to.
  if (const auto server = config.server_by_vbucket(vbucket, replica + 1);
      server.has_value() && server.value() < config.nodes.size()) {
    return server;
  }
  return {};
};
```

Here the index survives the resolver, and the timeout is produced downstream of it:

1. `VBucket.LocateReplica` indexes the endpoint list with it, catches the range error and returns `null`. `CouchbaseBucket.SendAsync` then fails `Nodes.TryGet` and throws `NodeNotAvailableException`.
2. That exception implements `IRetryable`, and `RetryReason.NodeNotAvailable` is not in `RetryReasonExtensions.AlwaysRetry()`, so `RetryOrchestrator` hands the decision to the operation's retry strategy.
3. `BestEffortRetryStrategy`, the default, retries any idempotent request with backoff. `ReplicaRead` is read-only, so it qualifies.

Nothing in that loop can change the node list, so it repeats until the deadline. Under `FailFastRetryStrategy` the third step returns `RetryAction.Duration(null)` and `NodeNotAvailableException` surfaces instead.

The two entries that both mean "no replica readable here" are reported differently. In the same class, whose `TestBucket` node list carries four nodes, `0` to `3`, so index `9` is outside it:

```csharp
[Fact]
public async Task An_Unassigned_Replica_Is_A_Replica_Index_Error()
{
    var bucket = new TestBucket();
    await bucket.ApplyVBucketMapAsync([1, -1], numReplicas: 1);

    await Assert.ThrowsAsync<ReplicaIndexCurrentlyUnavailableException>(() =>
        bucket.SendAsync(Key, GetReplicaStrategy.FromIndex(ReplicaIndex.First)));
    Assert.Empty(bucket.ServersThatReceivedOps);
}

[Fact]
public async Task A_Replica_On_A_Node_The_Topology_Does_Not_List_Is_A_Retryable_Transport_Error()
{
    var bucket = new TestBucket();
    await bucket.ApplyVBucketMapAsync([1, 9], numReplicas: 1);

    var exception = await Assert.ThrowsAsync<NodeNotAvailableException>(() =>
        bucket.SendAsync(Key, GetReplicaStrategy.FromIndex(ReplicaIndex.First)));

    Assert.IsAssignableFrom<IRetryable>(exception);
    Assert.Equal(RetryReason.NodeNotAvailable, exception.ResolveRetryReason());
    Assert.Empty(bucket.ServersThatReceivedOps);
}
```

src/Couchbase/MemcachedBucket.cs
```diff
@@ -112,6 +113,11 @@ internal override Task<ResponseStatus> SendAsync(IOperation op, CancellationToke
                 throw new InvalidOperationException("Bucket is not bootstrapped.");
             }
 
+            if (op.ReplicaStrategy != null)
+            {
+                throw new ReplicaIndexOutOfBoundsException($"Bucket {Name} is a memcached bucket and has no replicas.");
+            }
```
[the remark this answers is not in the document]
emilienbev: No need to add it for memcached buckets

emilienbev:
@avsej 
> A vBucket map row can name a node index that the node list does not contain. [RFC-0053 revision 23](https://github.com/couchbaselabs/sdk-rfcs/blob/36f7d3f3b8ca3a514fe04173afb7e3342055d106/rfc/0053-sdk3-crud.md?plain=1#L710-L720) gives decision rules for a -1 entry and says nothing about this one, so the two SDKs are free to differ, and they do.

Added that, might be worth raising in the RFC as well

src/Couchbase/KeyValue/Options.cs
```diff
@@ -441,6 +442,136 @@ public record ReadOnly(
...
+        /// <summary>
+        /// Inject a custom <see cref="IRetryStrategy"/>.
+        /// </summary>
+        /// <param name="retryStrategy"></param>
+        /// <returns>A <see cref="GetReplicaOptions"/> instance for chaining.</returns>
+        public GetReplicaOptions RetryStrategy(IRetryStrategy retryStrategy)
+        {
+            RetryStrategyValue = retryStrategy;
+            return this;
+        }
```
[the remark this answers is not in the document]
emilienbev: This is outside the scope of this PR and is more of a design opinion. Refusing this here

src/Couchbase/CouchbaseBucket.cs
```diff
@@ -305,6 +306,31 @@ async Task<IViewResult<TKey, TValue>> Func()
...
+                    BucketName = Name,
+                    ClientContextId = op.Opaque.ToStringInvariant(),
+                    DocumentKey = op.Key,
+                    CollectionName = op.CName,
+                    ScopeName = op.SName,
+                    OpCode = op.OpCode,
+                    DispatchedFrom = op.LastDispatchedFrom,
+                    DispatchedTo = op.LastDispatchedTo,
+                    RetryReasons = op.RetryReasons
+                };
```
davidkelly: `Status` is never assigned, so it lands on `default(ResponseStatus)` — which is `Success = 0x0000`, not the `None = -1` unset sentinel. Since `KeyValueErrorContext.Status` is serialised through `CamelCaseStringEnumConverter`, every `ReplicaIndexOutOfBoundsException` and `ReplicaIndexCurrentlyUnavailableException` reports itself as a success. Running the real path in `CouchbaseBucketGetReplicaTests` gives:

```json
{"dispatchedFrom":null,"dispatchedTo":null,"documentKey":"key","clientContextId":"3","cas":0,
 "status":"success","bucketName":"default","collectionName":null,"scopeName":null,
 "message":null,"opCode":"replicaRead","retryReasons":[]}
```

`Status = ResponseStatus.None` fixes it, and `Unavailable_Replica_Without_Wrap_Throws_And_Sends_Nothing` is the natural place to assert it alongside the `DocumentKey`/`BucketName`/`OpCode` checks it already makes.

Separately, worth asking whether this should build its own context at all rather than reusing `RetryOrchestrator.CreateKeyValueErrorContext`. The two have already drifted: that one also sets `Cas` and `Message`, and it reads `operation.ClientContextId` — which `OperationBase` hardcodes to `null`, so the `op.Opaque.ToStringInvariant()` here is actually the better behaviour of the two. But the result is that KV error contexts are now inconsistent with each other depending on which path produced them.

src/Couchbase/KeyValue/CouchbaseCollection.cs
```diff
@@ -1009,6 +1009,37 @@ public async Task<IGetReplicaResult> GetAnyReplicaAsync(string id, GetAnyReplica
...
 
+        /// <inheritdoc />
+        public async Task<IGetReplicaResult> GetReplicaAsync(string id, GetReplicaStrategy strategy, GetReplicaOptions? options = null)
+        {
+            _bucket.ThrowIfBootStrapFailed();
+
+            if (strategy == null)
+            {
+                throw new ArgumentNullException(nameof(strategy));
+            }
```
davidkelly: This should be `InvalidArgumentException`, not `ArgumentNullException`.

The convention in this file is that public-API argument validation raises `InvalidArgumentException` — the `content is null` checks at lines 327, 353 and 554, and the spec-count check at line 593, all do. `ArgumentNullException` appears only in the constructor (lines 61-70), for DI-injected dependencies, which is a different class of failure: a programming error in wiring rather than a bad value from a user call.

It also disagrees with its own sibling. `GetReplicaStrategy.FromIndex` rejects an out-of-range `ReplicaIndex` via `ThrowHelper.ThrowInvalidArgumentException`, so the two ways a caller can get the strategy wrong raise two unrelated exception types.

The practical consequence is in FIT. `ErrorsUtil.ConvertException` is declared as `ConvertException(CouchbaseException err)` and maps `InvalidArgumentException` to `SdkInvalidArgumentException` on line 28. `ArgumentNullException` is not a `CouchbaseException`, so it can never enter that function — if the driver adds a null-strategy case, the other SDKs will report their invalid-argument type and .NET won't be able to. Same for any user who catches `CouchbaseException` around a KV call.

src/Couchbase/MemcachedBucket.cs
davidkelly: The exception type reads fine — a memcached bucket has no replicas at any index, so out-of-bounds is honest, and the RFC doesn't cover memcached either way. But this one goes out bare while the couchbase path takes trouble to decorate: `CouchbaseBucket.SelectReplica` catches the same exception type and attaches a `KeyValueErrorContext` with the document key, bucket, scope/collection, opcode and retry reasons. So the same exception class reaches the user in two different shapes depending on which bucket they happened to open, and on the memcached side `Context` is null.

This path is reachable — `MemcachedBucket` has `Scope`/`IScopeFactory`, so `CouchbaseCollection.GetReplicaAsync` runs against it and lands here.

Worth either attaching a context here too, or factoring the decoration out of `CouchbaseBucket.SelectReplica` so both bucket types share it.

davidkelly (changes requested):
3 small comments.   I think `InvalidArgumentException` makes sense to throw vs `ArgumentNullException`, similarly the `ResponseStatus.None` seems small and worth doing.   The 3rd one is completely your call, as is the details with that `ResponseStatus`.

PR #188, opened 2026-09-17
NCBC-4247: Fix non-atomic double-checked dispose race in ClusterContext

jeffrymorris:
Motivation
----------
Dispose() guarded re-entrancy with a plain bool check-then-set; concurrent callers could both pass it and race on tearing down ownedObjects, risking "Collection was modified" crashes

Modifications
-------------
- Lock it the same way Cluster.Dispose() already does
- Add a regression test that reproduces the race pre-fix and passes reliably post-fix

Change-Id: I64ab095208e96df78994e916fb0afcce4b04d968


tests/Couchbase.UnitTests/Core/ClusterContextTests.cs
```diff
@@ -426,6 +426,156 @@ public void When_Tracing_Enabled_Custom_To_CustomTraceListener_Not_Disposed()
...
+                {
+                    try { context.Dispose(); } catch (Exception ex) { secondException = ex; }
+                }) { IsBackground = true };
+                second.Start();
+
+                // Race the two possible, mutually exclusive, independently observable outcomes -
+                // do not treat a timeout as evidence of exclusion either way.
+                var signaledIndex = WaitHandle.WaitAny(
+                    new WaitHandle[] { secondContendedLock.WaitHandle, secondReachedInnerHook.WaitHandle },
+                    TimeSpan.FromSeconds(5));
```
davidkelly, line 535: Worth saying first that the important design calls here are right: treating a timeout as *inconclusive and therefore a failure* is correct, and using dedicated `Thread`s rather than `Task.Run` sidesteps threadpool-injection stalls, which is the usual way a test like this goes bad.

The remaining concern is the four wall-clock deadlines - this one, `firstEntered.Wait` at 516, and the two `Join`s at 563/566. On an oversubscribed GitHub Actions runner, a wall-clock deadline is the one thing we've repeatedly watched turn green tests red.

The useful observation is that none of these timeouts adds detection power. Under the fixed code the second caller is *guaranteed* to set `secondContendedLock`; under the old check-then-set code it is *guaranteed* to set `secondReachedInnerHook`. Either way the `WaitAny` returns within microseconds of the thread being scheduled. So the 5s can't distinguish a product defect from a slow runner - it exists only as a backstop for a third outcome this wait doesn't observe: the thread exiting without reaching either hook.

Observe that outcome explicitly and the clock becomes unnecessary. `firstEntered` is only ever a one-shot signal, so it can become a `ManualResetEventSlim` and expose `.WaitHandle` directly (`SemaphoreSlim` only offers `AvailableWaitHandle`):

```csharp
var firstEntered  = new ManualResetEventSlim(false);
var firstExited   = new ManualResetEventSlim(false);
var secondExited  = new ManualResetEventSlim(false);

var first = new Thread(() =>
{
    try { context.Dispose(); } catch (Exception ex) { firstException = ex; }
    finally { firstExited.Set(); }
}) { IsBackground = true };
```

then, with no timeout anywhere:

```csharp
var i = WaitHandle.WaitAny(new[] { firstEntered.WaitHandle, firstExited.WaitHandle });
Assert.True(i == 0, "The first Dispose() call returned without ever reaching DisposeTestHook.");

var j = WaitHandle.WaitAny(new[]
    { secondContendedLock.WaitHandle, secondReachedInnerHook.WaitHandle, secondExited.WaitHandle });
Assert.False(j == 1, "A second concurrent Dispose() reached the same disposal critical section ...");
Assert.False(j == 2, "The second Dispose() call returned without ever contending on the lock ...");
```

and plain `Join()` in the `finally`, since both threads have been released by that point.

Liveness then depends only on "each thread eventually runs to completion", which holds at any level of oversubscription: slowness can no longer produce a red build, and each case the timeout used to lump together as "inconclusive" now fails with a specific message. A genuine product deadlock is still caught - CI runs `--blame-hang --blame-hang-timeout 5min --blame-hang-dump-type full` (`build-and-test.yml:58`, `nightly-unit-tests.yml:131`), so that surfaces as a red plus a full dump.

tests/Couchbase.UnitTests/Core/ClusterContextTests.cs
```diff
@@ -426,6 +426,156 @@ public void When_Tracing_Enabled_Custom_To_CustomTraceListener_Not_Disposed()
...
+            // ThresholdTraceListener to _ownedObjects, so the teardown under test is the real one,
+            // not a no-op over an empty list.
+            options.WithThresholdTracing(new ThresholdOptions { Enabled = true });
+
+            var context = new ClusterContext(Mock.Of<ICluster>(), new CancellationTokenSource(), options);
+            context.Start();
+
+            var firstEntered = new SemaphoreSlim(0);
+            var releaseFirst = new SemaphoreSlim(0);
+            var secondReachedInnerHook = new ManualResetEventSlim(false);
```
davidkelly, line 476: These are never disposed, and line 534 touches `.WaitHandle` on both `ManualResetEventSlim`s, which forces each to allocate a real kernel `ManualResetEvent` - so every run leaks two OS handles on top of the managed objects. The rest of this file consistently uses `using`, so this looks like an oversight rather than a choice.

`using var` on each covers it. If the no-timeout change lands, `firstEntered` becomes a `ManualResetEventSlim` and there are two more to include:

```csharp
using var firstEntered = new ManualResetEventSlim(false);
using var releaseFirst = new SemaphoreSlim(0);
using var secondReachedInnerHook = new ManualResetEventSlim(false);
using var firstExited = new ManualResetEventSlim(false);
using var secondExited = new ManualResetEventSlim(false);
```

`secondContendedLock` at 522 wants the same treatment - it's declared inside the `try`, but a `using var` there is fine since both threads are joined in the `finally` before any of these go out of scope.

tests/Couchbase.UnitTests/Core/ClusterContextTests.cs
```diff
@@ -426,6 +426,156 @@ public void When_Tracing_Enabled_Custom_To_CustomTraceListener_Not_Disposed()
...
+                    "own non-blocking probe) for this to count as a pass.");
+
+                Assert.Equal(1, hookEntries);
+            }
+            finally
+            {
+                // Always let go of whichever caller(s) are parked in the hook. Without this, a
+                // failed assertion above would leave the first thread blocked in
+                // releaseFirst.Wait() forever - and since these are ordinary (non-background, but
+                // see IsBackground above) threads, that can hang the whole test process rather
```
davidkelly, line 559: This describes the threads as "ordinary (non-background...)", but both are created with `IsBackground = true` (505 and 528), so the hazard it warns about - a parked thread hanging the whole test process - can't occur. Looks like it was written against an earlier revision and will mislead the next reader. Suggest trimming to just the reason the release is needed:

```csharp
// Always let go of whichever caller(s) are parked in the hook, so a failed
// assertion above doesn't leave a thread blocked in releaseFirst.Wait().
// Releasing more than needed is harmless.
```

davidkelly (changes requested):
A few minor comments - otherwise LGTM.

src/Couchbase/Core/ClusterContext.cs
```diff
@@ -1051,26 +1074,46 @@ public async Task ConfigUpdatedAsync(BucketConfig globalConfig)
...
-            _disposed = true;
-            _configHandler?.Dispose();
-            _semaphore.Dispose();
-            _tokenSource?.Dispose();
 
-            foreach (var ownedObject in _ownedObjects)
+            // Equivalent to `lock (_disposeLock)`, expanded so a failed non-blocking probe can be
+            // observed (via DisposeLockContendedHook) in the same statement that discovers the
+            // lock is held - see the remarks on that hook for why that matters for testability.
+            var lockTaken = Monitor.TryEnter(_disposeLock, 0);
```
emilienbev, line 1081: > Equivalent to `lock (_disposeLock)`

I don't think that's true, that `Monitor.TryEnter` should be inside the try as well ( https://learn.microsoft.com/en-us/dotnet/api/system.threading.monitor.tryenter?view=net-10.0#remarks).

An aborted thread in between a successful `Monitor.TryEnter` and the following if statement would leak the lock.

src/Couchbase/Core/ClusterContext.cs
```diff
@@ -45,6 +45,29 @@ internal class ClusterContext : IDisposable, IConfigUpdateEventSink
         private readonly CancellationTokenSource _tokenSource;
         protected readonly ConcurrentDictionary<string, BucketBase> Buckets = new();
         private bool _disposed;
+        private readonly object _disposeLock = new();
+
+        // Test-only seam. Invoked once per call that passes the inner not-yet-disposed check,
```
emilienbev, line 50: The 18 lines of comment on those two properties are mostly about justifying the tests, I think we should bring them down to a single line and if really worth it move the details to the tests themselves.

tests/Couchbase.UnitTests/Core/ClusterContextTests.cs
```diff
@@ -426,6 +426,156 @@ public void When_Tracing_Enabled_Custom_To_CustomTraceListener_Not_Disposed()
             Assert.False(listener.Disposed);
         }
 
+        /// <summary>
+        /// Regression test: Dispose() used to guard re-entrancy with a plain, non-atomic
```
emilienbev, line 430: Same here about the wordy comments, we should try to make them more concise. 
This one recounts a bunch of previous iterations as well which are now gone, we don't need the history here.

src/Couchbase/Core/ClusterContext.cs
```diff
@@ -1051,26 +1074,46 @@ public async Task ConfigUpdatedAsync(BucketConfig globalConfig)
         public void Dispose()
         {
             if (_disposed) return;
```
emilienbev, line 1076: Actually now looking at it, why not use
`if (Interlocked.Exchange(ref _disposed, 1) == 1) return;` ? 
The manual Monitor lock below matters only if a caller does something after `Dispose()` that depends on teardown being complete (and right now `_disposed` is only read inside `Dispose()` itself, from `Cluster.Dispose()` which has its own lock.)

emilienbev (changes requested):

src/Couchbase/Core/ClusterContext.cs
jeffrymorris: done

jeffrymorris:
Going to redo this entirely.

PR #189, opened 2026-09-18
NCBC-4216: Use server-presented intermediate certificates during TLS validation

emilienbev:
## Motivation

The SDK rejected valid server certificates whenever the node's certificate was issued by an intermediate CA and the client trusted only the root (only nodes with a root-signed leaf passed chain validation).
This is because the old callback discarded server-sent intermediate CAs, and therefore nodes with an intermediate-signed leaf cert failed to validate the chain.

This PR replaces the validation callback with one that never modifies the chain .NET hands it, honours the intermediates the server sends, and adds real TLS handshake tests (by creating an actual PKI chain).

### Behaviour after this change

1. .NET validates the chain against the OS trust store and checks the host name. If it reports no error, the certificate is accepted.
2. If the only error is that the chain does not reach an OS trusted root, the SDK builds its own chain with the configured trust anchors (the certificates given through `WithTrustedServerCertificates`, or the bundled Capella CA when none are given) plus the intermediates the server sent. The certificate is accepted only if that chain builds and passes expiry, path length and usage checks. Server-sent certificates are never treated as trust anchors.
3. Host name mismatch, a missing server certificate, or any other error rejects. `Kv/HttpIgnoreRemoteCertificateNameMismatch` skips the host name check only, never the chain check.
4. The host name is also checked explicitly on .NET 7 and later. Windows does not report a name mismatch when the chain is also untrusted, which is the normal case for a private or Capella CA, so the platform verdict alone would let a certificate for another host through.

### Windows and .NET Framework

The `netstandard2.0` and `netstandard2.1` builds **cannot use custom trust anchors or the explicit host name check**, since those APIs arrived in .NET 5 and .NET 7. On those targets a private or Capella CA must be added to the Windows trust store. The OS then trusts the chain by itself and reports host name mismatches correctly. This is unchanged from before.

## Changes
- Replace `CallbackCreator` and the validators inside `CertificateFactory` with `ServerCertificateValidator` behind the existing `ICertificateValidationCallbackFactory`. 
- Rebuild the chain on an SDK owned `X509Chain` with copies of the anchors and the server-sent intermediates. The chain from `SslStream` is left untouched since .NET owns that chain and disposes what is in it after the callback.
- Return false instead of throwing when the server presents no certificate.
- Add an explicit RFC 6125 host name check on .NET 7 and later.
- Add a loopback TLS test harness that creates a throwaway PKI and drives the real callback through `SslStream`. Covers wire intermediates, two-level chains, the NCBC-4275 same-name subordinate CA, trust anchor enforcement, rogue wire roots, expiry, path length, host name checks, wildcard certificates, the Capella defaults and certificate ownership across repeated handshakes.
- Make the test server work on Windows by persisting generated server keys for SChannel.


tests/Couchbase.UnitTests/Core/IO/Authentication/X509/Helpers/TlsTestPki.cs
```diff
@@ -0,0 +1,169 @@
...
+    public static bool IsSelfSigned(X509Certificate2 cert)
+    {
+        if (!cert.SubjectName.RawData.AsSpan().SequenceEqual(cert.IssuerName.RawData))
+        {
+            return false;
+        }
+
+        var authorityKey = cert.Extensions.OfType<X509AuthorityKeyIdentifierExtension>().FirstOrDefault()?.KeyIdentifier;
+        var subjectKey = cert.Extensions.OfType<X509SubjectKeyIdentifierExtension>().FirstOrDefault()?.SubjectKeyIdentifierBytes;
+        return authorityKey is null || subjectKey is null || authorityKey.Value.Span.SequenceEqual(subjectKey.Value.Span);
```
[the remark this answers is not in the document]
emilienbev: Wrong, came with .NET 7. But I'll change the pre-processor directive to `NET7_OR_GREATER` to make it strictly accurate.

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
```diff
@@ -0,0 +1,238 @@
...
+            var reportedNameMismatch = errors != sslPolicyErrors;
+
+            if (_ignoreNameMismatch)
+            {
+                if (reportedNameMismatch)
+                {
+                    _logger.LogDebug("X509 ignoring certificate name mismatch");
+                }
+            }
+            else if (reportedNameMismatch || !MatchesTargetHost(sender, certificate))
```
davidkelly, line 81: **Please gate this on the platform verdict.**

`MatchesTargetHost` runs here even when the platform reported `SslPolicyErrors.None`. The check is strictly additive in the reject direction — a platform-reported mismatch already returns false on this same branch — so it can never accept anything new, and every divergence from the platform is a potential regression rather than a fix.

I measured `X509Certificate2.MatchesHostname` against the cases where a second implementation could disagree:

| case | result |
|---|---|
| case-insensitive, trailing dot, wildcard | match |
| IPv4 SAN / IPv6 SAN | match |
| CN fallback, no SAN | match |
| partial wildcard `w*.example.com` | **no match** |
| IP address in a `dNSName` SAN (`DNS:10.1.2.3`) | **no match** |
| IDN unicode target vs punycode SAN | **no match** |

Better than I feared, but the IP-in-a-`dNSName` row will bite: it's a common `openssl.cnf` mistake that other platforms match anyway, and connecting to on-prem nodes by IP is normal. That combination works today.

The Windows hole you're closing is specifically "chain untrusted **and** wrong host -> Windows reports only `ChainErrors`", which always has `errors != None`. So gating loses nothing:

```csharp
else if (reportedNameMismatch)
{
    _logger.LogInformation("X509 certificate name does not match the target host, rejecting");
    return false;
}

if (errors == SslPolicyErrors.None)
{
    return true;                      // platform validated the host name; don't second-guess it
}

if (errors != SslPolicyErrors.RemoteCertificateChainErrors) { ... }

// Chain errors: Windows may have suppressed the name mismatch, so check explicitly.
if (!_ignoreNameMismatch && !MatchesTargetHost(sender, certificate))
{
    return false;
}
```

Separately, line 83's message carries no target host, no subject and no SANs. This will present in the field as "TLS stopped working after upgrading" and that line gives support nothing — worth adding the redacted target host and cert subject. It matters more than usual because the only escape hatch is `IgnoreRemoteCertificateNameMismatch`, which disables host name verification entirely, including the Windows protection this PR adds.

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                _logger.LogInformation("X509 certificate name does not match the target host, rejecting");
+                return false;
+            }
+
+            if (errors == SslPolicyErrors.None)
```
davidkelly, line 87: **Design question, entirely your call — I don't think it belongs in this PR.**

This returns `true` before `_trustedCertificates` is ever read, so `WithTrustedServerCertificates(myPrivateCa)` doesn't pin: a certificate for the right host signed by *any* CA in the OS trust store is accepted. On a managed enterprise machine that includes the corporate TLS-inspection root.

Inherited from master, not introduced here. But `NoErrors_Accepts_WithoutBuildingAChain` now enshrines it with an unrelated anchor and a comment saying the configured anchors "are not consulted at all", which turns a carry-over into documented intent.

Java diverges. `core-io/.../env/SecurityConfig.java:282`:

```java
userSpecifiedTrustSource = builder.trustCertificates != null || builder.trustManagerFactory != null;
trustCertificates = userSpecifiedTrustSource ? builder.trustCertificates : defaultCaCertificates();
```

`defaultCaCertificates()` is Capella CA **plus the JVM trust store**, documented as what's trusted *"if no other trust source is specified"*. So the two SDKs agree on the default and diverge only once the user configures one.

The same split maps onto us cleanly: keep the short-circuit on defaults, skip it when `TrustedServerCertificateFactory != null`. The factory already knows which case it's in, so it's one `bool` on the constructor and one condition here. Breaks nobody today, and only tightens the path where someone explicitly asked for pinning.

Worth its own ticket rather than this PR — it's a security-semantics change wanting its own release note and FIT run.

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+#else
+        private bool MatchesTargetHost(object sender, X509Certificate certificate) => true;
+#endif
+
+#if !NET5_0_OR_GREATER
+        private bool ChainsToTrustedCertificate(X509Certificate certificate, X509Chain chain)
+        {
+            _logger.LogWarning(
```
davidkelly, line 134: Two small things on the netstandard path.

1. The description says this is "unchanged from before", but master's pass 1 (`chain.Build` against the system store with the configured certs in `ExtraStore`) sat **outside** the `#if NET5_0_OR_GREATER`, so it ran on .NET Framework too. `WithTrustedServerCertificates` is now a complete no-op on these targets — this method returns false without ever reading `_trustedCertificates`. Worth correcting in the description, since it's the sentence a future maintainer will trust.

2. This fires once per handshake, so a cluster-wide TLS outage on .NET Framework repeats it per connection attempt. The wording is good and actionable — an `Interlocked` guard to emit it once per validator instance would keep it useful without the volume.

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
```diff
@@ -0,0 +1,238 @@
...
+            // the runtime holds can be disposed by this method.
+            var ownedCertificates = new List<X509Certificate2>();
+            using var ownChain = new X509Chain();
+            try
+            {
+                var policy = ownChain.ChainPolicy;
+                policy.TrustMode = X509ChainTrustMode.CustomRootTrust;
+                policy.RevocationMode = _revocationMode;
+                policy.RevocationFlag = X509RevocationFlag.ExcludeRoot;
+                policy.ApplicationPolicy.Add(new Oid(ServerAuthOid));
```
davidkelly, line 151: **I think this tightening is right and should stay** — requiring `serverAuth` on a certificate used for server authentication is correct and matches what browsers do. But it has no equivalent anywhere in master and isn't listed among the behaviour changes, so it needs to be an explicit decision.

Measured on **macOS**, so this is not a Windows CryptoAPI quirk:

| leaf EKU | issuing CA EKU | master | this PR |
|---|---|---|---|
| `serverAuth` | none | pass | pass |
| *(no EKU extension)* | none | pass | pass |
| `clientAuth` only | none | pass | **fail `NotValidForUsage`** |
| `serverAuth` + `clientAuth` | none | pass | pass |
| `serverAuth` | `clientAuth` only | pass | **fail `NotValidForUsage`** (EKU nesting) |

Two asks:

- **A release note.** The realistic break is certificate reuse in an mTLS setup — one certificate for both the node and client roles, tagged `clientAuth` only. Correct for the client role, and it now fails the server role.
- **Tests.** The suite structurally can't catch either failing row: `TlsTestPki.CreateServerLeaf` always adds `serverAuth` and `CreateCa` adds no EKU at all. An optional `ekus` parameter on both, plus three cases (no-EKU accepts, clientAuth-only rejects, constrained-intermediate rejects), would pin it.

Was this deliberate, or did it come along with the rewrite? The answer decides whether the note reads as a hardening or a fix.

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+                // SslStream places the intermediates the server sent in the ExtraStore of the chain it
+                // built. They are needed to reach the anchor but are never trusted on their own.
+                foreach (var presented in chain.ChainPolicy.ExtraStore)
+                {
+                    policy.ExtraStore.Add(Copy(presented, ownedCertificates));
+                }
+
+                var leaf = certificate as X509Certificate2 ?? Copy(certificate, ownedCertificates);
```
davidkelly, line 165: **Could we copy the leaf too?**

The comment at line 141 says "Everything placed in the stores of this chain is a copy owned here". True for `CustomTrustStore` and `ExtraStore`, but not for the leaf — `certificate as X509Certificate2` aliases the instance `SslStream` owns, and the `finally` below disposes everything in `ownChain.ChainElements`.

I measured it and **it's safe today**: `ReferenceEquals` is false for every chain element against the passed-in leaf, on the success path and both failure paths, and a shared bundle survives repeated builds. So this is hardening, not a bug.

But it rests on an undocumented property of the chain engine returning distinct managed instances. If that ever changed, the symptom is exactly NCBC-4120 — a disposed certificate breaking every later handshake. On a PR about certificate ownership I'd rather not lean on it:

```csharp
var leaf = Copy(certificate, ownedCertificates);
```

One extra copy per handshake, and the comment becomes true by construction.

To be clear, the `finally` loop itself is **not** redundant — I checked, and `X509Chain.Dispose()` does not dispose the element certificates on .NET 8. Removing it would leak. It should stay.

src/Couchbase/Core/IO/Authentication/CertificateValidationCallbackFactory.cs
```diff
@@ -27,60 +29,31 @@ public CertificateValidationCallbackFactory(
...
+                return userCallback;
             }
 
-            // If custom callback is provided, use it
-            if (_tlsSettings.HttpCertificateValidationCallback != null)
-            {
-                return _tlsSettings.HttpCertificateValidationCallback;
-            }
+            // Resolved on every call so a trusted certificate factory that rotates is honoured by new connections.
+            var trustedCertificates = _tlsSettings.TrustedServerCertificateFactory?.GetCertificates()
```
davidkelly, line 46: Deleting the old pass 1 removed one capability worth documenting: it built against the **OS trust store** with the configured certificates in `ExtraStore`, so a user-supplied cert could complete a path to an OS-trusted root. Under `CustomRootTrust` the OS store is ignored, so that no longer works.

Measured:

| `CustomTrustStore` | `ExtraStore` | result |
|---|---|---|
| `{ICA}` | `{}` | fail `PartialChain` |
| `{Root, ICA}` | `{}` | **pass** |
| `{Root}` | `{ICA}` | pass — this PR's fix working |
| `{ICA}` | `{ICA}` | fail `PartialChain` |

Note row 2: `CustomTrustStore` entries participate in path building, not just anchoring, so there's nothing to fix in the code — supplying the root, or root + intermediate, works fine. The narrow case that regressed is root-in-OS-store + node omits the intermediate + user supplies *only* the intermediate.

Narrow enough that documenting seems proportionate. `ClusterOptions.WithTrustedServerCertificates` (`ClusterOptions.cs:398`) currently says only "The server CA certificates to trust", which is exactly the ambiguity:

```csharp
/// <remarks>
/// <para>
/// The collection must contain the self-signed root of each chain you intend to trust.
/// Intermediate CAs may also be included and will be used to complete a path, but a
/// certificate that is not self-signed cannot act as a trust anchor on its own.
/// </para>
/// <para>
/// When the platform's own validation fails, the SDK rebuilds the chain against these
/// certificates alone. The OS trust store is not consulted for that rebuild, so a root
/// present only in the OS trust store will not be found. Intermediates presented by the
/// server are used to complete the chain but are never treated as trust anchors.
/// </para>
/// <para>
/// On the <c>netstandard2.0</c> and <c>netstandard2.1</c> builds these certificates are
/// not used: custom trust anchors require .NET 5 or later. Add the CA to the OS trust
/// store instead, which also lets the platform report host name mismatches correctly.
/// </para>
/// </remarks>
```

tests/Couchbase.UnitTests/Core/IO/Authentication/ServerCertificateValidatorTests.cs
```diff
@@ -0,0 +1,518 @@
...
+
+        var accepted = validator.Validate(new object(), null, null, SslPolicyErrors.RemoteCertificateNotAvailable);
+
+        Assert.False(accepted, "A handshake with no server certificate must be rejected.");
+    }
+
+    [Fact]
+    public void NoErrors_Accepts_WithoutBuildingAChain()
+    {
+        // The OS trust store already approved the chain, so the configured anchors are not consulted at all.
```
davidkelly, line 387: Could we soften this comment? The behaviour is real and worth pinning, but "the configured anchors are not consulted at all" reads as settled design rather than an inherited carry-over — see my note on `ServerCertificateValidator.cs:87`. Something that records the behaviour and points at the open question would stop the next person reading it as intentional.

davidkelly (changes requested):
Ok - Claude noted a few things.   I probed it to be more thorough and sure, and it made the various comments.   I figured it would be better to keep its comments in here as they will have more context for your Claude session.   I'll mark as request changes,  and you can take/not take the ones that sound right.   Overall this looks good to me!

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
emilienbev: Agreed and done, check now runs only on the chain error path, and I added a test that simulates the Windows case. The IP in a dNSName case only works on Windows anyway. On Linux and macOS .NET matches an IP against iPAddress SANs only, and Java does the same. The log now has the redacted subject and target host.

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
emilienbev: Agreed fixed 1), but left the warning as is. `CreateForKv` builds a new validator per connection, so a guard per instance wouldn't reduce the log amount. Every failed handshake already logs its reason on each attempt as well

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
emilienbev: Agreed it's out of scope, I think we should evaluate this one a bit more. Keeping the OS' trust store for now so users who pass certs but hit a publicly trusted node don't break.
I'll open a ticket for it. Cert pinning also can't work on `netstandard` since it doesn't have support for custom trust anchors (source of a bunch of our convoluted logic)

src/Couchbase/Core/IO/Authentication/ServerCertificateValidator.cs
emilienbev: Yes this is a hardening but thanks for pointing it out because I _think_ this can be optional sometimes. I need to check. 
Before the authenticator refactor there was a single `ICertificateFactory` where both server CAs and client certificates would go in and we'd have some logic to try and separate them (which didn't rely on just the `serverAuth`/`clientAuth` field I believe). Then with the authenticator refactor I separated them with the server certificates going into `TlsSettings` and client ones in the `CertificateAuthenticator`. I'll get back to this one.

PR #190, opened 2026-09-21
NCBC-4304: single query options are dropped by reflection and SetAuthenticator casts to Cluster

emilienbev:
## Motivation

Fix two .NET FIT performer bugs.

1. `SingleQueryConfigTest.adhocFalse` fails because the single query path mapped the driver's query options onto the SDK by reflection to save verbose code but that's very error-prone. 
The proto property is `Adhoc` and the SDK method is `AdHoc`, so the option was dropped. `adhocTrue` only passed because true is the default.

2. `SetAuthenticatorTest.canSetAuthenticator` fails on CNG because the performer cast the cluster to the classic `Cluster` type before calling `Authenticator`. On a couchbase2 connection the cluster is a `StellarCluster`, so the cast throws.

This PR maps each query option explicitly and calls `Authenticator` through the `ICluster` interface.

## Changes
- Replace the reflection block with an explicit mapper from the proto query options to `TransactionQueryOptions`. Options the .NET API cannot carry now raise `NotSupportedException` instead of being ignored.
- Remove the `Cluster` casts for password, certificate and JWT authenticators.


davidkelly (approved):
👍 LGTM

PR #192, opened 2026-09-22
NCBC-4303: mTLS rotation doesn't swap the cached certificates

emilienbev:
## Motivation

mTLS certificate rotation didn't swap the cached client certificates. The refresh stored the new collection with `Interlocked.Exchange` and then wrote its return value (the old collection) back into the cache. `HasUpdates` still reported true, so the HTTP handler was recreated and new KV connections opened on every tick, all with the stale certificates. 

Two smaller bugs sat in the same class:
- The expiry was parsed from a culture formatted date string, which threw on day first locales and silently skipped the refresh. 
- `Dispose` tested the disposed flag before setting it, so the refresh timer was never stopped.

This PR assigns the new collection directly, reads the expiry from `NotAfter` and makes `Dispose` stop the timer.

## Changes
- Replace the `Interlocked.Exchange` write-back with a plain assignment in both rotating factories. Every access already runs under the lock.
- Read the certificate expiry from `NotAfter` instead of parsing `GetExpirationDateString()`.
- Fix the inverted guard in `Dispose`, dispose the timer under the refresh lock and clear it.
- Tests assert the cache contents after a refresh, that a repeat refresh reports no updates, and that `Dispose` stops the refreshes.


src/Couchbase/Core/IO/Authentication/X509/RotatingCertificateFactory.cs
```diff
@@ -80,9 +79,7 @@ public void RefreshCertificates(object state)
 
                 if (validNewCertificates.Count > 0)
                 {
-                    _cachedCertificates =
-                        Interlocked.Exchange(ref _cachedCertificates,
-                            validNewCertificates);
+                    _cachedCertificates = validNewCertificates;
```
davidkelly: **The cache now holds only the newly-seen certificates, which drops still-valid ones.**

This is a consequence of the swap actually taking effect rather than a pre-existing issue, which is why I'm raising it here. Before this fix, the `Interlocked.Exchange` write-back left `_cachedCertificates` holding the previous *full* collection. Now it holds `validNewCertificates`, which by construction contains only the certificates that were **not already cached**.

During a normal rotation overlap that loses certificates the factory is still returning:

- cache is `[certA]`
- rotation begins; the factory now returns `[certA (still valid, 60 days left), certB (new)]`
- `certA` is filtered out by `!_cachedCertificates.Contains(certificate)` because it *is* already cached
- `validNewCertificates` is `[certB]`, so the cache becomes `[certB]` alone

`certA` is no longer presented at the handshake even though the factory is still offering it. If the server's CA hint matches `certA`, client auth fails on every new KV connection and every recreated HTTP handler - which is exactly when this code path runs.

The smallest fix keeps `validNewCertificates` purely as the change signal and caches the full usable set:

```csharp
var validNewCertificates = new X509Certificate2Collection();  // change detection only
var usableCertificates = new X509Certificate2Collection();    // what we actually cache
var possibleNewCertificates = _certificateFactoryImplementation.GetCertificates();
foreach (var certificate in possibleNewCertificates)
{
    if (certificate.NotAfter - DateTime.Today <= _expiresIn)
    {
        continue;
    }

    usableCertificates.Add(certificate);
    if (!_cachedCertificates.Contains(certificate))
    {
        validNewCertificates.Add(certificate);
    }
}

if (validNewCertificates.Count > 0)
{
    _cachedCertificates = usableCertificates;
    _hasChanges = true;
}
```

`DelegatingCertificateFactory.cs:41` has the same two lines and the same issue.

On the existing tests: `RefreshClientHandler_WithValidNewCertificates_ShouldUpdateCacheAndSetHasUpdates` uses a `SetupSequence` of two disjoint collections, so the factory stops returning the old certificate and `Assert.DoesNotContain(initialCertificates[0], refreshed)` is correct for that scenario. The overlap case - the factory returning `[old, new]` together - isn't covered, which is why this slipped through. A test with an overlapping second collection asserting **both** certificates survive would pin it.

src/Couchbase/Core/IO/Authentication/X509/RotatingCertificateFactory.cs
```diff
@@ -105,8 +102,14 @@ public void Dispose()
     {
         if (_disposed)
         {
-            _disposed = true;
+            return;
+        }
+
+        _disposed = true;
+        lock (_syncObj)
```
davidkelly: Note rather than a change request - two things about the new disposal behaviour that I think are worth deciding together. Raising them here because "Dispose stops the refreshes" is a guarantee this PR newly makes and tests for.

**1. `Dispose()` can now block for as long as a certificate fetch takes.** `RefreshCertificates` holds this same lock from `:53` through `:98`, which spans the `_certificateFactoryImplementation.GetCertificates()` call at `:70` - a user-supplied implementation that may hit disk, a network certificate store or a key vault, with no timeout under our control. The old `Dispose` didn't take the lock, so this is new exposure. It needs a slow fetch to coincide with shutdown, so it's unlikely - but shutdown is where hangs get noticed and blamed.

**2. `GetCertificates()` can resurrect the timer afterwards.** `_disposed` is set here but never read anywhere else, and `GetCertificates()` at `:34` creates a new timer whenever `_cachedCertificates.Count == 0`. A call after disposal on an empty cache therefore builds a fresh timer that nothing will ever dispose, and the refresh loop runs for the life of the process.

Worth being precise that the lock does **not** prevent this - it only serializes against a *concurrent* `GetCertificates`. Resurrection is purely about ordering: once `Dispose` has released the lock, a later `GetCertificates` acquires it freely and recreates the timer.

Reachability caveat, in fairness: this needs the cache to be empty, which means the underlying factory returned an empty collection at least once. On the normal path the cache is populated and `GetCertificates()` just returns it, so disposal sticks.

**The two interact.** A `_disposed` check in `GetCertificates()` is the more fundamental of the two, and having it weakens the case for the lock - `GetCertificates` is the only thing that ever *creates* a timer, and `RefreshCertificates`'s re-arm at `:96` already no-ops once `_timer` is null. So a `_disposed` guard plus a lock-free `Dispose` would cover both; keeping the lock and adding the guard would too. Entirely your call which - I just didn't want either to be a side effect rather than a decision.

src/Couchbase/Core/IO/Authentication/X509/RotatingCertificateFactory.cs
davidkelly, line 82: **The cache now holds only the newly-seen certificates, which drops still-valid ones.**

This is a consequence of the swap actually taking effect rather than a pre-existing issue, which is why I'm raising it here. Before this fix, the `Interlocked.Exchange` write-back left `_cachedCertificates` holding the previous *full* collection. Now it holds `validNewCertificates`, which by construction contains only the certificates that were **not already cached**.

During a normal rotation overlap that loses certificates the factory is still returning:

- cache is `[certA]`
- rotation begins; the factory now returns `[certA (still valid, 60 days left), certB (new)]`
- `certA` is filtered out by `!_cachedCertificates.Contains(certificate)` because it *is* already cached
- `validNewCertificates` is `[certB]`, so the cache becomes `[certB]` alone

`certA` is no longer presented at the handshake even though the factory is still offering it. If the server's CA hint matches `certA`, client auth fails on every new KV connection and every recreated HTTP handler - which is exactly when this code path runs.

The smallest fix keeps `validNewCertificates` purely as the change signal and caches the full usable set:

```csharp
var validNewCertificates = new X509Certificate2Collection();  // change detection only
var usableCertificates = new X509Certificate2Collection();    // what we actually cache
var possibleNewCertificates = _certificateFactoryImplementation.GetCertificates();
foreach (var certificate in possibleNewCertificates)
{
    if (certificate.NotAfter - DateTime.Today <= _expiresIn)
    {
        continue;
    }

    usableCertificates.Add(certificate);
    if (!_cachedCertificates.Contains(certificate))
    {
        validNewCertificates.Add(certificate);
    }
}

if (validNewCertificates.Count > 0)
{
    _cachedCertificates = usableCertificates;
    _hasChanges = true;
}
```

`DelegatingCertificateFactory.cs:41` has the same two lines and the same issue.

On the existing tests: `RefreshClientHandler_WithValidNewCertificates_ShouldUpdateCacheAndSetHasUpdates` uses a `SetupSequence` of two disjoint collections, so the factory stops returning the old certificate and `Assert.DoesNotContain(initialCertificates[0], refreshed)` is correct for that scenario. The overlap case - the factory returning `[old, new]` together - isn't covered, which is why this slipped through. A test with an overlapping second collection asserting **both** certificates survive would pin it.

src/Couchbase/Core/IO/Authentication/X509/RotatingCertificateFactory.cs
davidkelly, line 109: Note rather than a change request - two things about the new disposal behaviour that I think are worth deciding together. Raising them here because "Dispose stops the refreshes" is a guarantee this PR newly makes and tests for.

**1. `Dispose()` can now block for as long as a certificate fetch takes.** `RefreshCertificates` holds this same lock from `:53` through `:98`, which spans the `_certificateFactoryImplementation.GetCertificates()` call at `:70` - a user-supplied implementation that may hit disk, a network certificate store or a key vault, with no timeout under our control. The old `Dispose` didn't take the lock, so this is new exposure. It needs a slow fetch to coincide with shutdown, so it's unlikely - but shutdown is where hangs get noticed and blamed.

**2. `GetCertificates()` can resurrect the timer afterwards.** `_disposed` is set here but never read anywhere else, and `GetCertificates()` at `:34` creates a new timer whenever `_cachedCertificates.Count == 0`. A call after disposal on an empty cache therefore builds a fresh timer that nothing will ever dispose, and the refresh loop runs for the life of the process.

Worth being precise that the lock does **not** prevent this - it only serializes against a *concurrent* `GetCertificates`. Resurrection is purely about ordering: once `Dispose` has released the lock, a later `GetCertificates` acquires it freely and recreates the timer.

Reachability caveat, in fairness: this needs the cache to be empty, which means the underlying factory returned an empty collection at least once. On the normal path the cache is populated and `GetCertificates()` just returns it, so disposal sticks.

**The two interact.** A `_disposed` check in `GetCertificates()` is the more fundamental of the two, and having it weakens the case for the lock - `GetCertificates` is the only thing that ever *creates* a timer, and `RefreshCertificates`'s re-arm at `:96` already no-ops once `_timer` is null. So a `_disposed` guard plus a lock-free `Dispose` would cover both; keeping the lock and adding the guard would too. Entirely your call which - I just didn't want either to be a side effect rather than a decision.

tests/Couchbase.UnitTests/Core/IO/Authentication/X509/RotatingCertificateFactoryTests.cs
```diff
@@ -300,6 +307,73 @@ public void RefreshClientHandler_WithNullTimer_ShouldReturnEarly()
...
+        // Act
+        var result = factory.GetCertificates();
+
+        var completedTask = await Task.WhenAny(calledTwiceTcs.Task, Task.Delay(TimeSpan.FromSeconds(30)));
+        Assert.True(completedTask == calledTwiceTcs.Task, "Timer should have fired at least once before Dispose");
+
+        factory.Dispose();
+        var countAfterDispose = Volatile.Read(ref callCount);
+
+        await Task.Delay(200);
```
davidkelly, line 340: **This test is timing-dependent in a way that will flake on the Windows CI leg.**

```csharp
factory.Dispose();
var countAfterDispose = Volatile.Read(ref callCount);
await Task.Delay(200);
factory.RefreshCertificates(factory);
Assert.Equal(countAfterDispose, Volatile.Read(ref callCount));
```

The `calledTwiceTcs` signal earlier in the test is exactly right. The `Task.Delay(200)` against a 20 ms timer isn't - on a loaded agent it passes vacuously, and a callback already in flight when `Dispose` lands can flake it.

It also isn't buying coverage: the assertion is already guaranteed by the `_timer == null` early return at `RotatingCertificateFactory.cs:56`, because `Dispose` takes `_syncObj` and therefore waits out any in-flight refresh. Deleting the line leaves the test no weaker.

The deterministic version is cheap and follows the pattern NCBC-4233 already established here. `TimerFactory.CreateWithFlowSuppressed(TimeProvider, ...)` is already in master, `Microsoft.Bcl.TimeProvider` is referenced unconditionally so it works on `netstandard2.0`, and `FakeTimeProvider` is used in five other unit test files.

The only wrinkle is that `RotatingCertificateFactory` is public and constructed only by users, so adding an optional `TimeProvider` parameter to the existing constructor would be a binary break. An internal overload avoids that entirely, and `InternalsVisibleTo("Couchbase.UnitTests")` makes it directly usable from the test:

```csharp
public RotatingCertificateFactory(
    ICertificateFactory certificateFactoryImplementation, TimeSpan interval, TimeSpan expiresIn, ILogger? logger = null)
    : this(certificateFactoryImplementation, interval, expiresIn, logger, TimeProvider.System)
{
}

internal RotatingCertificateFactory(
    ICertificateFactory certificateFactoryImplementation, TimeSpan interval, TimeSpan expiresIn,
    ILogger? logger, TimeProvider timeProvider)
```

Then `_timer` becomes `ITimer?`, `:43` passes `_timeProvider`, and `:62` uses `Timeout.InfiniteTimeSpan` since `ITimer.Change` has no `int` overload. The test drives `FakeTimeProvider.Advance` instead of waiting, and the 30 s `Task.Delay` guard at `:334` goes away too.

davidkelly (commented):
Ok this looks good, with a few comments.   Claude noted a pre-existing issue this code didn't fix, and probably shouldn't.   Instead of expanding scope, I wanted Claude to just mention it for you to decide about.   I'm ok adding to this, or making a ticket just for that.   I am trying to stop expanding the scope constantly like Claude is likely to do...   So up to you - it is mentioned below in Claude's version of the review summary below:

=====
The three fixes are correct and worth landing. I checked each: the `Interlocked.Exchange` write-back really was self-cancelling (it put the old collection straight back), `NotAfter` is the right replacement for the culture-parsed expiry string with no semantic drift since both are local time, and the `Dispose` guard really was inverted so the timer was never stopped.

My comments are scoped to what these changes introduce or activate rather than to pre-existing problems in these classes. One of them I'd treat as blocking — making the swap actually happen means the cache is now replaced by only the newly-seen certificates, so a certificate the factory is still returning gets dropped during a rotation overlap. The other two are a note on the new disposal behavior and a flaky wall-clock wait in the new test.

One thing left open, for a follow-up rather than this PR. The description says the bug caused the HTTP handler to be recreated on every tick.  This fixes the certificate-staleness half of that, but the `HasUpdates` half is untouched: nothing ever consumes the flag. `CouchbaseHttpClientFactory.Create()` checks it on every HTTP request and RecreateHandler disposes the in-use handler, but the flag is only ever cleared at the start of the next RefreshCertificates. That leaves two opposite failure modes from the same root cause — the handler is ebuilt on every request for a full refresh interval after each rotation, and conversely, if two ticks pass with no `Create()` between them the flag goes true then false unobserved and the rotated certificates are silently never picked up. `DelegatingCertificateFactory` is the worse case: `GetCertificates()` sets the flag on first populate and that class has no timer, so unless the user calls `RefreshCertificates` themselves the flag is true forever and the handler is torn down on every single request. All of that predates this PR and none of it is in this diff, so I'd rather see it as its own ticket — but it does mean the symptom in the description isn't fully closed, and it'd be worth softening that wording so the next person reading it isn't misled.

src/Couchbase/Core/IO/Authentication/X509/RotatingCertificateFactory.cs
emilienbev: Yes agreed, pushed a commit to fix this. The cache now mirrors whatever the user's factory returns _less_ the certs close to expiry (overlaps keep both and removed certs get dropped), stops the cache flipping between two certs on every refresh (and keeps the current certs with a warning if nothing usable comes back)

src/Couchbase/Core/IO/Authentication/X509/RotatingCertificateFactory.cs
emilienbev: Added the disposed guard as well, and `GetCertificates()` now only starts a timer when there is none.
I kept the lock in `Dispose()`, because without it a running refresh can call `Change` on the disposed timer _outside_ the catch, and break the timer thread.

tests/Couchbase.UnitTests/Core/IO/Authentication/X509/RotatingCertificateFactoryTests.cs
emilienbev: Yeah actually the delay isn't doing anything, I removed it (and there wasn't a need for a `FakeTimeProvider` here). The direct `RefreshCertificates()` call after `Dispose()` is what catches the bug. The old `Dispose()` didn't clear the timer, so the test failed every time without any waiting.

